Issue #19, opened 2017-11-15
EventDefinition / EventReference not taken into account


iamwyza:
For some reason MonoMod is changing how events are being migrated slightly.  I don't know that it actually changes their functionality.  But it seems to take something that is fairly simple and makes it exceptionally verbose.

```csharp
        [HookInfo("Called at the end of the attack function", "HeroController.Attack")]
        public event AfterAttackHandler AfterAttackHook;

        public void AfterAttack(AttackDirection dir)
        {
            AfterAttackHook?.Invoke(dir);
        }

```
is instead moved into the new assembly as:
```csharp
[MonoModAdded]
public void AfterAttack(AttackDirection dir)
{
	AfterAttackHandler afterAttackHook = this.AfterAttackHook;
	if (afterAttackHook == null)
	{
		return;
	}
	afterAttackHook(dir);
}
[MonoModAdded]
[CompilerGenerated]
public void add_AfterAttackHook(AfterAttackHandler value)
{
	AfterAttackHandler afterAttackHandler = this.AfterAttackHook;
	AfterAttackHandler afterAttackHandler2;
	do
	{
		afterAttackHandler2 = afterAttackHandler;
		AfterAttackHandler value2 = (AfterAttackHandler)Delegate.Combine(afterAttackHandler2, value);
		afterAttackHandler = Interlocked.CompareExchange<AfterAttackHandler>(ref this.AfterAttackHook, value2, afterAttackHandler2);
	}
	while (afterAttackHandler != afterAttackHandler2);
}
[MonoModAdded]
[CompilerGenerated]
public void remove_AfterAttackHook(AfterAttackHandler value)
{
	AfterAttackHandler afterAttackHandler = this.AfterAttackHook;
	AfterAttackHandler afterAttackHandler2;
	do
	{
		afterAttackHandler2 = afterAttackHandler;
		AfterAttackHandler value2 = (AfterAttackHandler)Delegate.Remove(afterAttackHandler2, value);
		afterAttackHandler = Interlocked.CompareExchange<AfterAttackHandler>(ref this.AfterAttackHook, value2, afterAttackHandler2);
	}
	while (afterAttackHandler != afterAttackHandler2);
}
```

0x0ade:
That's #14 - I prefer your issue description, so I'll just close that one and rename this one. Thanks for giving an example 👍 

[...]

PR #20, opened 2017-11-17
Attempt to fix EventDefinition and EventReference

iamwyza:
Initial attempt at fixing #19.  Seems to work when the target type doesn't have an existing event, not sure if it will work on overriding.  Other thing that might be broken is attribute duplication.  Not sure why, but it doesn't seem to break anything.

### DIFF
--- a/MonoMod/MonoModder.cs
+++ b/MonoMod/MonoModder.cs
@@ -1296,8 +1296,9 @@ namespace MonoMod {
                 (addMethod = PatchMethod(targetType, adder)) != null)
             {
                 targetEvent.AddMethod = addMethod;
+                /* Old code that would add the base attributes to the add method.  Causes weird behavior.
                 foreach (CustomAttribute attrib in srcEvent.CustomAttributes)
-                    addMethod.CustomAttributes.Add(attrib.Clone());
+                    addMethod.CustomAttributes.Add(attrib.Clone()); */
                 propMethods?.Add(adder);
             }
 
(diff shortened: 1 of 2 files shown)


0x0ade:
Looks great! 👍 

Custom attributes being carried over from the mod's property / event definition to the original one is actually a "feature." Some mods apply custom attributes to the original definition that way.

I'd just merge it immediately, but it sounds like you're not done yet?

iamwyza:
Well I don't know how to fix the duplicate attribute issue.  Basically when it comes over it does this:
(where HookInfo is an attribute)
```csharp
[HookInfo("Some stuff about this thing", "Something Else")]
[method:HookInfo("Some stuff about this thing", "Something Else")]
public event ThingHandler ThingHook;
```

it should only have 1 [HookInfo].

MonoMod/MonoModder.cs
```diff
@@ -1212,6 +1216,110 @@ public virtual TypeReference FindType(string name, bool runtimeName)
...
+            foreach (CustomAttribute attrib in srcEvent.CustomAttributes)
+                targetEvent.CustomAttributes.Add(attrib.Clone());
+
+            MethodDefinition adder = srcEvent.AddMethod;
+            if (adder != null &&
+                (addMethod = PatchMethod(targetType, adder)) != null)
+            {
+                targetEvent.AddMethod = addMethod;
+                foreach (CustomAttribute attrib in srcEvent.CustomAttributes)
+                    addMethod.CustomAttributes.Add(attrib.Clone());
```
0x0ade: @iamwyza I added those two lines for the get, set and other methods in the properties patcher. Someone requested for MonoMod to process attributes added to the property on its "children," too, but I can think of another way to solve that.

Feel free to remove those here and further down :)

iamwyza: Ok.  I'll do that when i get off work, and re-test.  If it works I'll push those changes.  It makes sense as to why it was doing that now that I realize what the reasoning was for it.  It may indeed need to be handled differently, but for now, just cloning the attributes at the top level of the event rather than the backing methods should be sufficient.

PR #22, opened 2017-11-21
Fixes #21

### LINKED ISSUE #21: Duplicate assembly references of differing versions
My target assembly has a reference to mscorlib 2.0.5.0 which I believe is some subset of the Unity version of mscorlib.  I found a way to make my merging DLL to use that, but since MonoMod injects some types into the assembly, that assembly then gets a reference to the mscorlib used by MonoMod, which is mscorlib 2.0.0.0.  This causes weird behavior in unity.  So I then have to open up the assembly in dnspy and save it in order to get it to remove that extra reference.

The types copied from a source assembly should have the reference to mscorlib changed to whatever the target assembly's version is.

iamwyza:
The original fix in #21 was broken because the code that added the module to the cache would find "mscorlib" under it's simple name and assume that was the right one even when the fullname was different.  This now explicitly checks fullname first and only checks short name if fullname was not passed.  

Additionally the logic to determine which was the higher version meant that it never found duplicates if the 2nd version was lower than the 1st.  Now it always finds the highest, regardless of order.

Finally The relink has to happen after MonoModWasHere is added as MonoModWasHere also would reference the version of MonoMod.

### DIFF
--- a/MonoMod/MonoModder.cs
+++ b/MonoMod/MonoModder.cs
@@ -1,4 +1,4 @@
-﻿using Mono.Cecil;
+using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Cecil.Mdb;
 using Mono.Cecil.Pdb;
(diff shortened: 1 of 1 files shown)


0x0ade:
Thanks for the fix :)

PR #27, opened 2018-07-24
Fixed MonoModConstructor

seanpr96:
Dunno if there's a better util function to use instead of this substring stuff, but I needed this working


0x0ade:
Looks great! I'll need to check if it breaks other existing mods / mod loaders, though.

cc @ZandrXandr can you please test this PR against the games you're modding right now? Or at least hand out a test build of Partiality to a small group.

0x0ade:
I'll just merge this and force everyone to adapt to this instead.

PR #29, opened 2018-10-06
Fix behaviour of MoveAfterLabel

Chicken-Bones:


MonoMod.RuntimeDetour/HookGen/HookILCursor.cs
```diff
@@ -97,17 +92,17 @@ public HookILCursor Clone()
...
-
         public void MarkLabel(HookILLabel label) {
-            _LastLabel = label;
             label.Target = Next;
+            _insertAfterLabels = new [] { label };
         }
 
         public void MoveAfterLabel() {
-            _LastLabel = LabelNext;
+			_insertAfterLabels = HookIL._Labels.Where(l => l.Target == Next).ToArray();
```
0x0ade:
```cs
HookIL._Labels.Where(l => l.Target == Next)
```

This specifically shouldn't be necessary, as each `HookILCursor` just holds what `HookIL` context it belongs to and its `ID`.

Creating a new label checks if there is already an ID assigned pointing to the same target. Setting the target of one label with ID `N` "sets" the target of all labels with ID `N`.

Just wondering, did it not work properly?

0x0ade:
I'll fix some code style issues locally real quick - still at the gamejam

0x0ade:
Done, thanks again :)

PR #39, opened 2019-02-07
Improve aarch64 comments

leo60228:
Remove FIXME (due to it being fixed), and use `.+8` instead of `#8` (I'm unsure if they're equivalent, but `.+8` was used when assembling and `#8` reads as an absolute address).


0x0ade:
`#8` matches the other detours and is what [shell-storm's online assembler and disassembler](http://shell-storm.org/online/Online-Assembler-and-Disassembler/?opcodes=%22%5Cx4f%5Cx00%5Cx00%5Cx58%22&arch=arm64&endianness=little#disassembly) spits out after roundtripping. I'll merge your PR nevertheless.

PR #45, opened 2019-04-11
Free us from the shackles of DMD with EmitDelegate! Down with the oppressors

Chicken-Bones:


0x0ade:
17/10 too many generics

PR #46, opened 2019-04-18
Add FindDeep extension methods

opl-:
Let's add more ~~bloat~~ functionality! Implemented it so might as well make a PR. Feel free to reject it if you feel it shouldn't exist.


opl-:
I just noticed that the build failed and... What? I'm a little confused by this. How can `BaseType` be not found? It quite clearly exists in .NET 3.5 according to the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.type.basetype?view=netframework-3.5)

PR #62, opened 2020-02-04
Add MONOMOD_DEBUGIL_FORMAT=pdb/mdb env var

zatherz:
This PR adds checks for the `MONOMOD_DEBUGIL_FORMAT` env variable and `--pdb`/`--mdb` options to control the output of the DebugIL program. `MONOMOD_DEBUGIL_FORMAT` is case insensitive and may be `PDB` to force PDB output, `MDB` to force MDB output or anything else to default to automatic detection.


zatherz:
Closing (the scope is too small)

PR #67, opened 2020-06-01
.NET Core JIT Hooks

nike4613:
The main repo version of https://github.com/MonoMod/MonoMod.Common/pull/4.


nike4613:
Note that currently the hooks do not correctly identify .NET 5, so those tests fail

nike4613:
The behavior still seems inconsistent. I'm not sure what can be done to make it more consistent, short of delaying all re-JITing while patching methods. That may end up being the correct approach though.

MonoMod.RuntimeDetour/Detour.cs
```diff
@@ -385,10 +400,29 @@ public Detour(Expression<Action> from, Expression<Action> to)
                 return;
 
             // GetNativeStart to avoid repins and managed copies.
-            _TopDetour = new NativeDetour(Method.GetNativeStart(), TargetReal.GetNativeStart());
+            _TopDetour = new NativeDetour(Method.Pin().GetNativeStart(), TargetReal.Pin().GetNativeStart());
```
0x0ade, line 403: Method and TargetReal get pinned here for the top detour only but don't get unpinned until Free. I'm a little bit blind and worried right now so I'll better ask than be sorry afterwards: Is it supposed to be this way or are they supposed to be unpinned in TopUndo and only TopUndo instead?

MonoMod.RuntimeDetour/NativeDetour.cs
```diff
@@ -27,7 +27,8 @@ public unsafe class NativeDetour : IDetour {
         public bool IsValid { get; private set; }
         public bool IsApplied { get; private set; }
 
-        public readonly NativeDetourData Data;
+        private NativeDetourData _Data;
+        public NativeDetourData Data => _Data;
```
0x0ade, line 31: This is technically a breaking change, but realistically almost nobody should be using this directly, and whoever is using this will probably recompile with an update anyway, bypassing the break. Just writing this down here for future reference.

MonoMod.UnitTest/MonoMod.UnitTest.csproj
```diff
@@ -14,7 +14,7 @@
     <OutputType>Library</OutputType>
     <SignAssembly>skip</SignAssembly>
     <!-- FIXME: RE-ENABLE TESTS WITH TIERED COMPILATION -->
-    <TieredCompilation>false</TieredCompilation>
+    <!--<TieredCompilation>false</TieredCompilation>-->
```
0x0ade, line 17: The FIXME here seems to no longer be relevant as we are now testing with tiered compilation.

MonoMod.RuntimeDetour/Detour.cs
nike4613: To be honest, I'm not entirely sure. Would there be any major consequence to leaving methods pinned forever? I went ahead and added Unpin calls in TopUndo to be safe, though I have no idea what the correct thing to do would be.

0x0ade: Theoretically it would mostly affect Mono (untestable right now) and DynamicMethods - which if I'm being honest, the more I look at it and look back at my previous review, the less I see them being accounted for. Welp, I'll need to figure out what's going on with them before continuing to review this PR 😅

PR #104, opened 2022-09-09
Add simple implementation of HookEndpointManager

Chicken-Bones:
Keeps existing HookGen assemblies alive and happy. Stripped down a lot from the old HookEndpointManager to go with the new theme of minimal bloat.

Things other people may want (but tModLoader doesn't need):
- `OnAdd/Remove/Modify/Unmodify` events
- A way to enumerate active hooks
- A way to disable a subset of hooks (and not just `Clear`)

### DIFF
--- a/src/MonoMod.RuntimeDetour.New/HookGen/HookEndpointManager.cs
+++ b/src/MonoMod.RuntimeDetour.New/HookGen/HookEndpointManager.cs
@@ -1,0 +1,55 @@
+﻿using MonoMod.Cil;
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace MonoMod.RuntimeDetour.HookGen {
+
+    /// <summary>
+    /// Provided for back-compat with old versions of HookGen
+    /// </summary>
+    public static class HookEndpointManager {
+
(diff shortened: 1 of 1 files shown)


src/MonoMod.RuntimeDetour.New/HookGen/HookEndpointManager.cs
```diff
@@ -0,0 +1,55 @@
...
+        public static void Add<T>(MethodBase method, Delegate hookDelegate) where T : Delegate => Add(method, hookDelegate);
+        public static void Add(MethodBase method, Delegate hookDelegate) {
+            Hooks.Add((method, hookDelegate), new Hook(method, hookDelegate));
+        }
+
+        public static void Remove<T>(MethodBase method, Delegate hookDelegate) where T : Delegate => Remove(method, hookDelegate);
+        public static void Remove(MethodBase method, Delegate hookDelegate) {
+            if (Hooks.TryGetValue((method, hookDelegate), out var hook)) {
+                Hooks.Remove((method, hookDelegate));
+                hook.Undo();
```
nike4613, line 26: When removing these detours, it would be better to call `Dispose`.

nike4613 (changes requested):

PR #116, opened 2023-02-06
Fix custom attribute relinking + properly clone custom debug info

Popax21:
See https://github.com/MonoMod/MonoMod.Common/pull/32: this contains both the changes I made there + some additional ones I will open a PR for on the master branch here after it is merged. I've opened this PR because the changes fix issues which seem to be present in the reorganize branch as well.

### DIFF
--- a/src/MonoMod.Patcher/MonoModder.cs
+++ b/src/MonoMod.Patcher/MonoModder.cs
@@ -1594,7 +1594,7 @@ namespace MonoMod {
             for (int i = 0; i < type.GenericParameters.Count; i++) {
                 type.GenericParameters[i] = type.GenericParameters[i].Relink(Relinker, type);
                 for (int j = 0; j < type.GenericParameters[i].CustomAttributes.Count; j++)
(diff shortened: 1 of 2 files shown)


nike4613 (approved):
This largely looks fine.

I'm not especially familiar with the patcher, so I can't really comment on that (certainly doesn't help that there don't really exist tests for it), but I'd like to pose the same question I asked over on the Common PR: Is this method in Utils only used by the patcher? If not, where else is it used, and are we sure that this change won't cause any issues in those other locations?

Popax21:
> This largely looks fine.
> 
> I'm not especially familiar with the patcher, so I can't really comment on that (certainly doesn't help that there don't really exist tests for it), but I'd like to pose the same question I asked over on the Common PR: Is this method in Utils only used by the patcher? If not, where else is it used, and are we sure that this change won't cause any issues in those other locations?

I'm not particularly aware in which places this method can be called, but considering that this only changes the way debug symbols are cloned to now behave properly (instead of the earlier behavior, which was still referencing the incorrect `Instruction` objects), I don't see how this could cause anything to break.

nike4613:
Fair enough.

PR #117, opened 2023-02-07
Clean up compiler warnings / errors

Popax21:
This change fixes various compiler warnings which occur when targeting .NET 6.0/7.0 (the only target framework which seems to build at all) on Linux (they presumably appear on other platforms as well, but I have no way of testing this). It consists of:
- adding `#pragma warning` directives to various places
- removing unused imports
- adding entries to `NoWarn` when appropriate
- adding a missing XMLDoc comment to a public member of `ISystem`
- disabling code analyzers for `Patcher` and `HookGen` - to my knowledge, they are gonna be reworked soon, but I've still decided to temporarily disable them to prevent the build log from becoming cluttered
- upgrading the used `Microsoft.CodeAnalysis.NetAnalyzers` version (from 6.0.0 to 7.0.0)
- reworking the target frameworks of some projects (additionally, .NET 6.0 is moved to the front of the list - this is because OmniSharp (used by VS Code for Intelisense) uses the first entry of the list to determine which .NET version to use for a project, for which .NET 6.0 makes the most sense right now in my opinion)

I've tested both .NET 6.0 and 7.0 on Linux, both work / test fine after this patch is applied. However, it's probably best if someone can test all the other frameworks / platforms before merging.

### DIFF
--- a/src/MonoMod.FrameworkTests/MonoMod.FrameworkTests.csproj
+++ b/src/MonoMod.FrameworkTests/MonoMod.FrameworkTests.csproj
@@ -4,7 +4,8 @@
     <OutputType>Exe</OutputType>
     <IsPackable>false</IsPackable>
     <TargetFrameworks>$(TargetFrameworks);netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.4.0 And $(NETCoreAppMaximumVersion) &gt;= 7.0">$(TargetFrameworks);net7.0</TargetFrameworks>
+
+    <NoWarn>$(NoWarn);CA1303</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/MonoMod.FrameworkTests/Program.cs
+++ b/src/MonoMod.FrameworkTests/Program.cs
@@ -1,20 +1,13 @@
 ﻿#define MM
 
-using Mono.Cecil.Cil;
-using MonoMod.Cil;
-using MonoMod.Core;
-using MonoMod.Core.Platforms;
 using MonoMod.RuntimeDetour;
 
-using MonoMod.Backports;
 using System;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using MonoMod.Utils;
 using System.Runtime.InteropServices;
 
 #if NETCOREAPP1_0_OR_GREATER
-using Xunit;
 using Xunit.Abstractions;
 #endif
 
--- a/src/MonoMod.Patcher/MonoMod.Patcher.csproj
+++ b/src/MonoMod.Patcher/MonoMod.Patcher.csproj
@@ -7,7 +7,8 @@
     <Description>General purpose .NET assembly modding "basework". This package contains the core IL patcher and relinker.</Description>
     <PackageTags>$(PackageTags)</PackageTags>
     <Nullable>annotations</Nullable>
-    <TargetFrameworks>net452;netstandard2.0;net6.0</TargetFrameworks>
+    <MMUseCodeAnalyzers>false</MMUseCodeAnalyzers>
+    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild> <!-- Remove after / during rework -->
     
     <OutputType>Exe</OutputType>
   </PropertyGroup>
--- a/src/MonoMod.RuntimeDetour.HookGen/MonoMod.RuntimeDetour.HookGen.csproj
+++ b/src/MonoMod.RuntimeDetour.HookGen/MonoMod.RuntimeDetour.HookGen.csproj
@@ -7,8 +7,8 @@
     <PackageId>MonoMod.RuntimeDetour.HookGen</PackageId>
     <Description>Auto-generate hook helper .dlls, hook arbitrary methods via events: On.Namespace.Type.Method += YourHandlerHere;</Description>
     <PackageTags>RuntimeDetour;detour;detours;$(PackageTags)</PackageTags>
-    <TargetFrameworks>net452;netstandard2.0;net6.0</TargetFrameworks>
     <Nullable>annotations</Nullable>
+    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild> <!-- Remove after / during rework -->
 
     <OutputType>Exe</OutputType>
 
--- a/src/MonoMod.RuntimeDetour/DetourContext.cs
+++ b/src/MonoMod.RuntimeDetour/DetourContext.cs
@@ -77,6 +77,7 @@ namespace MonoMod.RuntimeDetour {
         /// which can be used in a <see langword="using"/> block to automatically pop it from the context stack.
         /// </summary>
         /// <returns>A <see cref="DataScope"/> which manages the lifetime of this context on the context stack.</returns>
+        [CLSCompliant(false)] //TODO: DataScope isn't CLS compliant - is this bad?
         public DataScope Use() => PushContext(this);
 
         /// <summary>
--- a/src/MonoMod.SourceGen.Internal/Utils/FastDelegateInvokersGenerator.cs
+++ b/src/MonoMod.SourceGen.Internal/Utils/FastDelegateInvokersGenerator.cs
@@ -79,6 +79,7 @@ namespace MonoMod.SourceGen.Internal.Utils {
             ctx.AddSource($"{methodName}.g.cs", sb.ToString());
         }
 
+#pragma warning disable CA1305 // Specify IFormatProvider
         private static void BuildSourceFor(CodeBuilder builder, GeneratorMethod method, out string methodName) {
             var methodSymbol = method.Method;
             var maxArgs = method.MaxArgs;
@@ -189,6 +190,7 @@ namespace MonoMod.SourceGen.Internal.Utils {
 
             ctx.AppendExitContext(builder);
         }
+#pragma warning restore CA1305 // Specify IFormatProvider
 
         private static string ComputeNameForIdx(int idx) {
             // this index is structured as follows (low to high bits)
--- a/src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
+++ b/src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
@@ -11,11 +11,13 @@
 
     <TargetFrameworks>net46;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
     <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 16.6.0 And $(NETCoreAppMaximumVersion) &gt;= 5.0">$(TargetFrameworks);net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">$(TargetFrameworks);net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">net6.0;$(TargetFrameworks)</TargetFrameworks>
     <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.4.0 And $(NETCoreAppMaximumVersion) &gt;= 7.0">$(TargetFrameworks);net7.0</TargetFrameworks>
     <OutputType>Library</OutputType>
     <SignAssembly>skip</SignAssembly>
     <Nullable>annotations</Nullable>
+
+    <NoWarn>$(NoWarn);CS0649;CA1812</NoWarn>
 
     <XunitVersion Condition="'$(XunitVersion)' == ''">2.4.2</XunitVersion>
     <XunitVsRunnerVersion>$(XunitVersion)</XunitVsRunnerVersion>
--- a/src/MonoMod.UnitTest/RuntimeDetour/DetourRedoTest.cs
+++ b/src/MonoMod.UnitTest/RuntimeDetour/DetourRedoTest.cs
@@ -4,15 +4,15 @@
 extern alias New;
 
 using Xunit;
-using New::MonoMod.RuntimeDetour;
-using System;
-using System.Reflection;
-using System.Runtime.CompilerServices;
-using MonoMod.Utils;
-using System.Reflection.Emit;
-using System.Text;
-using System.Collections.Generic;
-using System.Collections.Concurrent;
+// using New::MonoMod.RuntimeDetour;
+// using System;
+// using System.Reflection;
+// using System.Runtime.CompilerServices;
+// using MonoMod.Utils;
+// using System.Reflection.Emit;
+// using System.Text;
+// using System.Collections.Generic;
+// using System.Collections.Concurrent;
 using Xunit.Abstractions;
 
 namespace MonoMod.UnitTest {
--- a/src/MonoMod.UnitTest/RuntimeDetour/MultiHookTest.cs
+++ b/src/MonoMod.UnitTest/RuntimeDetour/MultiHookTest.cs
@@ -14,9 +14,11 @@ using Xunit.Abstractions;
 
 namespace MonoMod.UnitTest {
     public class ManualMultiHookTest : TestBase {
+#pragma warning disable CA2213 // Disposable fields should be disposed
         Hook h1;
         Hook h2;
         ILHook hIL;
+#pragma warning restore CA2213 // Disposable fields should be disposed
 
         private bool h1Run;
         private bool h2Run;
--- a/src/MonoMod.UnitTest/TestLogHelpers.cs
+++ b/src/MonoMod.UnitTest/TestLogHelpers.cs
@@ -19,7 +19,9 @@ namespace MonoMod.UnitTest {
                         return;
                 }
                 var localTime = time.ToLocalTime();
+#pragma warning disable CA1305 // Specify IFormatProvider
                 helper.WriteLine($"[{source}]({localTime}) {level.FastToString()}: {message}");
+#pragma warning restore CA1305 // Specify IFormatProvider
             };
         }
 
--- a/src/MonoMod.Utils/Logs/DebugLog.cs
+++ b/src/MonoMod.Utils/Logs/DebugLog.cs
@@ -149,12 +149,14 @@ namespace MonoMod.Logs {
             // we do this log here because we want to always log to the debugger when its attached, instead of just if its attached at startup
             if (Debugger.IsAttached) {
                 try {
+#pragma warning disable CA1305 // Specify IFormatProvider
                     // Even though Debugger.Log won't do anything when no debugger is attached, it's still worth guarding it with a check of Debugger.IsAttached for a few reasons:
                     //  1. It avoids the allocation of the message string when it wouldn't be used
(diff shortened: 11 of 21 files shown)


src/MonoMod.FrameworkTests/Program.cs
```diff
@@ -1,20 +1,13 @@
...
 using System;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using MonoMod.Utils;
 using System.Runtime.InteropServices;
 
 #if NETCOREAPP1_0_OR_GREATER
-using Xunit;
 using Xunit.Abstractions;
 #endif
```
nike4613, line 12: This is in a similar situation to the MacOS system implementation; all of these have been used at some point and will likely be used in the future. FrameworkTests is a project used exclusively for experimenting and testing different aspects of MonoMod on various platforms without dealing with the stuff Xunit runs before actual tests.

src/MonoMod.Patcher/MonoMod.Patcher.csproj
```diff
@@ -7,7 +7,8 @@
     <Description>General purpose .NET assembly modding "basework". This package contains the core IL patcher and relinker.</Description>
     <PackageTags>$(PackageTags)</PackageTags>
     <Nullable>annotations</Nullable>
-    <TargetFrameworks>net452;netstandard2.0;net6.0</TargetFrameworks>
```
nike4613, line 10: Should the patcher build for *all* of MonoMod's targets? I figured it was fine if it hit .NET Standard and one Framework and one Core, being that it's a standalone application first.

src/MonoMod.RuntimeDetour.HookGen/MonoMod.RuntimeDetour.HookGen.csproj
```diff
@@ -7,8 +7,8 @@
     <PackageId>MonoMod.RuntimeDetour.HookGen</PackageId>
     <Description>Auto-generate hook helper .dlls, hook arbitrary methods via events: On.Namespace.Type.Method += YourHandlerHere;</Description>
     <PackageTags>RuntimeDetour;detour;detours;$(PackageTags)</PackageTags>
-    <TargetFrameworks>net452;netstandard2.0;net6.0</TargetFrameworks>
```
nike4613, line 10: Same comment as for Patcher.

src/MonoMod.RuntimeDetour/DetourContext.cs
```diff
@@ -77,6 +77,7 @@ private sealed class ContextScopeHandler : ScopeHandlerBase {
         /// which can be used in a <see langword="using"/> block to automatically pop it from the context stack.
         /// </summary>
         /// <returns>A <see cref="DataScope"/> which manages the lifetime of this context on the context stack.</returns>
+        [CLSCompliant(false)] //TODO: DataScope isn't CLS compliant - is this bad?
         public DataScope Use() => PushContext(this);
```
nike4613, line 81: I left this one because I do intend to go through Utils and annotate CLS compliance at some point, I just haven't gotten around to it yet.

src/MonoMod.SourceGen.Internal/Utils/FastDelegateInvokersGenerator.cs
```diff
@@ -79,6 +79,7 @@ private sealed record GeneratorMethod(IMethodSymbol Method, SyntaxTokenList Modi
             ctx.AddSource($"{methodName}.g.cs", sb.ToString());
         }
 
+#pragma warning disable CA1305 // Specify IFormatProvider
```
nike4613, line 82: As mentioned before, I'd prefer `SuppressMessage` attributes, or in this case, explicitly passing `null` to the methods which take it.

src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
```diff
@@ -11,12 +11,14 @@
...
     <TargetFrameworks>net46;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
     <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 16.6.0 And $(NETCoreAppMaximumVersion) &gt;= 5.0">$(TargetFrameworks);net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">$(TargetFrameworks);net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">net6.0;$(TargetFrameworks)</TargetFrameworks>
     <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.4.0 And $(NETCoreAppMaximumVersion) &gt;= 7.0">$(TargetFrameworks);net7.0</TargetFrameworks>
     <OutputType>Library</OutputType>
     <SignAssembly>skip</SignAssembly>
     <Nullable>annotations</Nullable>
 
+    <NoWarn>$(NoWarn);CS0649;CA1812</NoWarn>
```
nike4613, line 20: Both of these warnings are things that I'd rather be explicitly suppressed where they exist.

src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
```diff
@@ -11,12 +11,14 @@
 
     <TargetFrameworks>net46;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
     <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 16.6.0 And $(NETCoreAppMaximumVersion) &gt;= 5.0">$(TargetFrameworks);net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">$(TargetFrameworks);net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">net6.0;$(TargetFrameworks)</TargetFrameworks>
```
nike4613, line 14: Should probably include a comment explaining why this is inconsistent. That being said, I don't think we support old SDKs anymore anyway, so all of these conditionals can probably be removed.

src/MonoMod.UnitTest/RuntimeDetour/DetourRedoTest.cs
```diff
@@ -4,15 +4,15 @@
...
-using System.Collections.Concurrent;
+// using New::MonoMod.RuntimeDetour;
+// using System;
+// using System.Reflection;
+// using System.Runtime.CompilerServices;
+// using MonoMod.Utils;
+// using System.Reflection.Emit;
+// using System.Text;
+// using System.Collections.Generic;
+// using System.Collections.Concurrent;
```
nike4613, line 15: This test needs to be rewritten entirely... I'd rather leave it as is until that happens, to serve as something of a reminder.

src/MonoMod.UnitTest/RuntimeDetour/MultiHookTest.cs
```diff
@@ -14,9 +14,11 @@
 
 namespace MonoMod.UnitTest {
     public class ManualMultiHookTest : TestBase {
+#pragma warning disable CA2213 // Disposable fields should be disposed
         Hook h1;
         Hook h2;
         ILHook hIL;
+#pragma warning restore CA2213 // Disposable fields should be disposed
```
nike4613, line 21: See earlier comments about suppressions.

src/MonoMod.UnitTest/TestLogHelpers.cs
```diff
@@ -19,7 +19,9 @@ internal static class TestLogHelpers {
                         return;
                 }
                 var localTime = time.ToLocalTime();
+#pragma warning disable CA1305 // Specify IFormatProvider
                 helper.WriteLine($"[{source}]({localTime}) {level.FastToString()}: {message}");
+#pragma warning restore CA1305 // Specify IFormatProvider
```
nike4613, line 24: Same as above. For this though, I'd rather just explicitly pass `null` to `FastToString`.

src/MonoMod.Utils/Logs/DebugLog.cs
```diff
@@ -149,12 +149,14 @@ private sealed class LogMessage {
...
             if (Debugger.IsAttached) {
                 try {
+#pragma warning disable CA1305 // Specify IFormatProvider
                     // Even though Debugger.Log won't do anything when no debugger is attached, it's still worth guarding it with a check of Debugger.IsAttached for a few reasons:
                     //  1. It avoids the allocation of the message string when it wouldn't be used
                     //  2. Debugger.Log is implemented as a QCall (on CoreCLR, and probably Framework) which pulls in all of the P/Invoke machinery, and necessitates a GC transition.
                     //     Debugger.IsAttached, on the other hand, is an FCall (MethodImplOptions.InternalCall) and likely elides the helper frames entirely, making it much faster.
                     Debugger.Log((int) message.Level, message.Source,
                         DebugFormatter.Format($"[{message.Source}] {message.Level.FastToString()}: {message.FormattedMessage}\n")); // the VS output window doesn't automatically add a newline
+#pragma warning restore CA1305 // Specify IFormatProvider
```
nike4613, line 159: Ditto.

src/MonoMod.Utils/Logs/DebugLog.cs
```diff
@@ -325,8 +327,11 @@ internal bool ShouldLevelRecordHoles(LogLevel level)
                             return;
                     }
 
+#pragma warning disable CA1305 // Specify IFormatProvider
                     var realTime = time.ToLocalTime();
                     var outMsg = $"[{source}]({realTime}) {level.FastToString()}: {msg}";
+#pragma warning restore CA1305 // Specify IFormatProvider
```
nike4613, line 333: Ditto.

tools/Common.CS.props
```diff
@@ -45,7 +45,7 @@
 
   <!-- For our C# projects, we want to pull in the code analyzers -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0">
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.0">
```
nike4613, line 48: I believe the reason that I hadn't upgraded this was because the new package doesn't actually work. I'd need to double check now, but the last couple times I did, the analyzers weren't actually loading.

tools/Common.NoTargets.props
```diff
@@ -2,7 +2,6 @@
 <Project>
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
```
nike4613, line 5: This should be kept because it fairly significantly improves build perf by only delegating to one TFM. Other than that, it doesn't matter what TFM it is here.

tools/Common.props
```diff
@@ -10,7 +10,7 @@
     <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
     
     <!-- All MonoMod projects (by default) target these frameworks -->
-    <TargetFrameworks>net35;net452;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net5.0;net7.0;netstandard2.0;net35;net452</TargetFrameworks>
```
nike4613, line 13: There should probably be a comment here explaining this ordering. I'd personally also prefer if it were strictly newest -> oldest, though that's not especially important.

nike4613 (changes requested):
Broadly speaking, most parts that still have warnings are either incomplete or I haven't gotten around to warning cleanup there. I've also noticed that different IDEs have a tendency to display different sets of warnings, (i.e. VS shows different warnings than Rider which shows different warnings from OmniSharp which shows different warnings than CLI) and notably VS doesn't show any warnings about `IFormatProvider` to me. I think I may have already disabled it in `.editorconfig`.

src/MonoMod.FrameworkTests/MonoMod.FrameworkTests.csproj
```diff
@@ -4,7 +4,8 @@
     <OutputType>Exe</OutputType>
     <IsPackable>false</IsPackable>
     <TargetFrameworks>$(TargetFrameworks);netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.4.0 And $(NETCoreAppMaximumVersion) &gt;= 7.0">$(TargetFrameworks);net7.0</TargetFrameworks>
+
+    <NoWarn>$(NoWarn);CA1303</NoWarn>
```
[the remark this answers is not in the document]
Popax21: It only affects this singular project, and going of the .editorconfig it being active in others is intentional.

src/MonoMod.Patcher/MonoMod.Patcher.csproj
Popax21: I figured if it was targeting basically the entire spectrum of .NET runtime kinds, it might as well target everything. I'll revert that change though as soon as I push my cleanup commit.

src/MonoMod.RuntimeDetour.HookGen/MonoMod.RuntimeDetour.HookGen.csproj
Popax21: > Same comment as for Patcher.

See response to Patcher as well.

src/MonoMod.RuntimeDetour/DetourContext.cs
Popax21: I'll remove the TODO + annotation in that case

src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
Popax21: Fixed - had to use pragmas for CS0649 though as it's a compiler warning.

tools/Common.CS.props
Popax21: At least on my machine, they work (at least during build time), and not using them causes a warning. I'll revert that though if it turns out to cause issues on other dev environments.

tools/Common.props
Popax21: I've put .NET 6.0 first considering it's an LTS version, and .NET 7.0 isn't fully supported yet from what I remember jade saying.

Popax21:
@nike4613 I've applied all the changes you've requested, please let me know if there's anything else I should change.

src/MonoMod.Patcher/MonoMod.Patcher.csproj
nike4613: I'm not sure what you mean. The patcher itself should support .NET 7 insofar as Cecil (and the patcher's own implementation) does, regardless of what TFM it's targeting. Plus, a .NET 6 application *should* be able to run on the .NET 7 runtime with minimal changes.

src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
nike4613: I believe you can `SuppressMessage` compiler built-in warnings, just not errors.

tools/Common.props
nike4613: .NET 7 should be fully supported, at least in RuntimeDetour. The only current limitation is around MacOS. Unless you're talking about support in, say, Rider, which I know is incomplete.

src/MonoMod.SourceGen.Internal/Utils/FastDelegateInvokersGenerator.cs
```diff
@@ -79,7 +80,7 @@ private sealed record GeneratorMethod(IMethodSymbol Method, SyntaxTokenList Modi
             ctx.AddSource($"{methodName}.g.cs", sb.ToString());
         }
 
-#pragma warning disable CA1305 // Specify IFormatProvider
+        [SuppressMessage("Globalization", "CA1305:Specify IFormatProvider", Justification="SourceGen is a standalone program, not using specific localization settings for integers isn't that important.")]
```
nike4613, line 83: Nit: this isn't strictly true. SourceGen is a Roslyn extension, not a standalone program.

Style nit: The `=` character for the property definition should have spaces around it, and `Justification` should be on a second line because the string is relatively long.

src/MonoMod.UnitTest/HookGen/HookGenRunTest.cs
```diff
@@ -63,6 +64,7 @@ public class HookGenRunTest : TestBase {
 }
 
 namespace MonoMod.UnitTest.HookGenTrash.Other {
+    [SuppressMessage("Performance", "CA1812:Avoid uninstantiated internal classes", Justification="Empty stub class.")]
```
nike4613, line 67: Ditto about the spaces, though the justifications here are short enough to stay on the same line.

src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
```diff
@@ -9,16 +9,11 @@
     <Description></Description>
     <PackageTags>$(PackageTags)</PackageTags>
 
-    <TargetFrameworks>net46;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 16.6.0 And $(NETCoreAppMaximumVersion) &gt;= 5.0">$(TargetFrameworks);net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.0.0 And $(NETCoreAppMaximumVersion) &gt;= 6.0">net6.0;$(TargetFrameworks)</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 17.4.0 And $(NETCoreAppMaximumVersion) &gt;= 7.0">$(TargetFrameworks);net7.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net5.0;net7.0;net46;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
```
nike4613, line 12: I still think we should keep TFMs in sorted order, whether ascending or descending.

src/MonoMod.UnitTest/RuntimeDetour/MultiHookTest.cs
```diff
@@ -11,14 +11,14 @@
 using Xunit;
 using Mono.Cecil.Cil;
 using Xunit.Abstractions;
+using System.Diagnostics.CodeAnalysis;
 
 namespace MonoMod.UnitTest {
+    [SuppressMessage("Usage", "CA2213:Disposable fields should be disposed", Justification = "Hooks are disposed by the test teardown method")]
```
nike4613, line 17: `Justification` should be on a new line.

nike4613 (commented):
I still need to double check that analysis works with the new analyzer versions.

Other than that, aside from minor style things, this should be good to go.

nike4613:
> I still need to double check that analysis works with the new analyzer versions.

I'm not able to reproduce the issue in my current environment (x86_64 Arch), need to try on Windows when I get home.

nike4613:
It seems to work on Windows too.

src/MonoMod.UnitTest/MonoMod.UnitTest.csproj
Popax21: I've kept .NET 6.0 as the first entry for the same reason as above, though I've added the same comment present in the `Common.props` file here as well now.

nike4613 (approved):

PR #118, opened 2023-02-08
Fix MonoModRulesModder for nested types

Popax21:
`MonoModRulesModder` is a subtype of `MonoModder` whose purpose is to bypass the relinker for the `MonoModRules` type when it's extracted into its own assembly to be executed as part of the patching process. However, the previous implementation didn't check if the referenced type is a nested type of `MonoModRules`, even though these types are also extracted. Usually, this isn't a problem, as the assembly the type is taken from is itself a mod assembly, for which the relinker uses `FindTypeDeep` to resolve types, which doesn't seem to cause any issues. However, when the `MonoModRules` type is parsed / executed explicitly, usually as part of a runtime relinker, this isn't the case, and as such the output assembly seems to be corrupted in a few subtle ways (presumably related to access modifiers), which can cause accesses to these nested types (including compiler generated types) to fail at runtime with a `TypeAccessException`/ `MemberAccessException`.

This very likely also affects the legacy codebase of MonoMod, but I've targeted this PR against the `reorganize` branch because this is the branch I'm currently in the process of switching to.

### DIFF
--- a/src/MonoMod.Patcher/InlineRT/MonoModRulesModder.cs
+++ b/src/MonoMod.Patcher/InlineRT/MonoModRulesModder.cs
@@ -1,5 +1,4 @@
-﻿using Mono.Cecil;
-using MonoMod.Utils;
+using Mono.Cecil;
 using System;
 
 namespace MonoMod.InlineRT {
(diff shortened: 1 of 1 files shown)


0x0ade:
I can't wait to eventually rewrite the entirety of the assembly patcher, hopefully sometime this year, if I find the spare time, as usual...

Thank you for this PR and the other one in the common repo :)

PR #126, opened 2023-03-05
Implement same-thread detour chain hotpatching support

Popax21:


src/MonoMod.RuntimeDetour/DetourManager.Native.cs
```diff
@@ -185,17 +194,44 @@ internal sealed class ChainDelegateState {
...
+                Helpers.Assert(dmd.Module is not null && dmd.Definition is not null);
+                var module = dmd.Module;
+
+                var il = dmd.GetILProcessor();
+
+                // instantiate a new System.InvalidOperationException and throw it
+                il.Emit(OpCodes.Ldstr, "Native detour has been removed");
+                il.Emit(OpCodes.Newobj, module.ImportReference(typeof(InvalidOperationException).GetConstructor(new Type[] { typeof(string) })));
+                il.Emit(OpCodes.Throw);
+
```
nike4613: Is there a reason to keep this separate from the removed stubs for managed methods? Could we not just pass the delegate's invoke signature to that to get a method which can be used for the target delegate?

src/MonoMod.RuntimeDetour/DetourManager.Shared.cs
```diff
@@ -190,10 +192,15 @@ internal sealed class DepGraph<TNode> {
...
 
-            public void WaitForChainUpdate() {
+            public bool WaitForChainUpdate() {
+                // If this is a nested call, allow the call to continue without waiting for the update to avoid deadlocks
+                if (Volatile.Read(ref ActiveCalls) > 1 && DetermineThreadCallDepth() > 1)
+                    return true;
+
                 _ = Interlocked.Decrement(ref ActiveCalls);
 
                 if (UpdatingThread == EnvironmentEx.CurrentManagedThreadId) {
```
nike4613, line 207: Isn't this now redundant?

src/MonoMod.RuntimeDetour/Hook.cs
```diff
@@ -454,8 +454,55 @@ public Hook(MethodBase source, Delegate target, IDetourFactory factory, DetourCo
         MethodInfo IDetour.InvokeTarget => realTarget;
 
 
-        private readonly MethodInfo trampoline;
-        MethodBase IDetour.NextTrampoline => trampoline;
+        private sealed class TrampolineData : IDetourTrampoline, IDisposable {
+
+            private readonly object LOCK = new object();
```
nike4613: I'm a little surprised that the editorconfig didn't cause a warning for this. Generally, I'd name it like a normal local (i.e. `syncRoot` or similar) or just lock on `this`, given that we know exactly who will lock on this type.

nike4613 (changes requested):

src/MonoMod.RuntimeDetour/DetourManager.Native.cs
Popax21: You're right, I'll move it into the shared DetourManager code.

src/MonoMod.RuntimeDetour/DetourManager.Shared.cs
Popax21: This code path is intended to catch the edge case where one thread calls the method, another thread tries to add/remove a hook, and then the first thread tries to perform a nested call - without this edge case this would end up deadlocking, as the nested call will block on the second thread, which will wait for the first thread's original call to return.

src/MonoMod.RuntimeDetour/Hook.cs
Popax21: Got it, will change that.

nike4613 (approved):

PR #137, opened 2023-07-04
Remove reference to removed BaseLoader from README

dhasenan:
According to Discord, BaseLoader was removed circa 2019. The README still mentions it. This can be confusing to newcomers.


nike4613:
`reorganize` already removed this reference, and is now the main branch.

PR #142, opened 2023-08-15
Add .NET 8 case for `DMDEmitDynamicMethodGenerator`

chairclr:
Preliminary support for .NET 8
The name of the internal runtime type field on `DynamicMethod` for .NET 8 is _runtimeType
https://source.dot.net/#System.Private.CoreLib/src/System/Reflection/Emit/DynamicMethod.CoreCLR.cs,20


nike4613:
Integrated with #150

PR #143, opened 2023-09-28
Add net8 support

jborean93:
Adds support for dotnet 8 by including the new JIT GUID and CorJit VTable information needed.


jborean93:
I never got back to this unfortunately, for some reason the hooks aren't applying in some cases but does in others. Feel free to use whatever is here if anyone needs a starting point for net8 support.

PR #153, opened 2023-11-28
Implement pipe-based page probing

Popax21:


nike4613:
One thought actually occurred to me the other day about this. Isn't this basically just [`IsBadReadPtr`](https://devblogs.microsoft.com/oldnewthing/20060927-07/?p=29563)? Will this not end up having the same problems?

PR #162, opened 2024-01-19
Improve ILCursor label positioning behavior.

Chicken-Bones:
A couple of oversights/bugs in the initial implementation:
- `ILCursor.MoveAfterLabels` doesn't move into EH ranges at all. This means that editing instructions at the start of an try range, or immediately after a handler requires manual updating of the references in `Body.ExceptionHandlers`.
- `ILCursor.Remove` loses label positioning. As an example, removing the first instruction in an `else` block will repoint the incoming branch, but the cursor would move to before the branch requiring a `.MoveAfterLabels()` call to emit a replacement to the removed instruction.
- `ILCursor.Remove` will leave dangling labels or exception ranges when called without a preceding `.MoveAfterLabels()` call.

It is unlikely that many people are relying on the current deficient behaviour. Anyone who was working around the existing issues by calling `.MoveAfterLabels()` after `Remove`, or manually updating EH targets is also unlikely to be affected, this just makes such workarounds unnecessary.

See someone working around the lack of EH support in https://github.com/WEGFan/Celeste-Mod-China-Mirror/blob/7c7834f2e30087b846135b3beba99c27ca2be3fb/Modules/ChinaMirror.cs#L282C22-L282C38

Overall this should make `ILCursor` a little more featureful and robust. 

Not included:
- Support for specifying _which_ try range to emit in, when there are multiple that start at the same instruction. 
  - Reason: Any user with enough knowledge to write code distinguishing two handlers with the same try range can surely update the `HandlerStart` values themselves.

### DIFF
--- a/src/MonoMod.UnitTest/RuntimeDetour/ILCursorTest.cs
+++ b/src/MonoMod.UnitTest/RuntimeDetour/ILCursorTest.cs
@@ -1,4 +1,4 @@
-﻿extern alias New;
+extern alias New;
 
 using MonoMod.Cil;
 using New::MonoMod.RuntimeDetour;
@@ -59,10 +59,19 @@ namespace MonoMod.UnitTest {
 
         [Fact]
         public void TestMoveBeforeTry() {
+            // moves before the try, but into the if block, and thus doesn't get executed
+            Assert.Equal("trywhenfinallyreturn",
+                TestTryCatchFinallyILEdit(c => {
+                    c.GotoNext(MoveType.Before, i => i.MatchLdloc(0), i => i.MatchLdstr("try"));
+                    EmitThrowEx(c);
+                })
+            );
+
             // Moves before the try-catch, so the exception is unhandled
             Assert.Throws<Exception>(() => {
                 TestTryCatchFinallyILEdit(c => {
-                    c.GotoNext(MoveType.Before, i => i.MatchLdloc(0));
+                    c.GotoNext(MoveType.Before, i => i.MatchLdloc(0), i => i.MatchLdstr("try"));
+                    c.MoveAfterLabels(intoEHRanges: false);
                     EmitThrowEx(c);
                 });
             });
@@ -73,7 +82,7 @@ namespace MonoMod.UnitTest {
             // Moves into the try-handler, triggering the general catch clause
             Assert.Equal("catchfinallyreturn", 
                 TestTryCatchFinallyILEdit(c => {
-                    c.GotoNext(MoveType.AfterLabel, i => i.MatchLdloc(0));
+                    c.GotoNext(MoveType.AfterLabel, i => i.MatchLdloc(0), i => i.MatchLdstr("try"));
                     EmitThrowEx(c);
                 })
             );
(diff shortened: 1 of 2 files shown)


nike4613:
Think you could add some unit tests to ensure correct behavior here? I'd like to start moving to having a more comprehensive test suite, and this seems like a simple enough thing to start with.

Chicken-Bones:
Sure, can you make me a test harness?

nike4613:
Can you not use xUnit in MonoMod.UnitTest for this? I think we already have some tests vaguely similar in nature for stuff like ILHook.

src/MonoMod.Utils/Cil/ILCursor.cs
```diff
@@ -460,40 +478,63 @@ public ILCursor GotoLabel(ILLabel label, MoveType moveType = MoveType.AfterLabel
...
+                    if (eh.FilterStart == Next) eh.FilterStart = instr;
+                }
+            }
+
+            if (_afterHandlerEnds) {
+                foreach (var eh in Body.ExceptionHandlers) {
+                    if (eh.TryEnd == Next) eh.TryEnd = instr;
+                    if (eh.HandlerEnd == Next) eh.HandlerEnd = instr;
+                }
+            }
```
nike4613, line 499: Isn't this logic backwards? If the current cursor is *after* EH end markers, then those end markers shouldn't be moved, right? They should only be moved when the cursor is placed *before* those end markers.

src/MonoMod.UnitTest/RuntimeDetour/ILCursorTest.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        private static void EmitAppendTest(ILCursor c) {
+            c.EmitLdloc(0);
+            c.EmitLdstr("TEST");
+            c.EmitCallvirt(new Func<string, StringBuilder>(new StringBuilder().Append).Method);
+            c.EmitPop();
+        }
+
+        private static void EmitThrowEx(ILCursor c) {
+            c.EmitNewobj(typeof(Exception).GetConstructors()[0]);
```
nike4613: Pretty sure it's not safe to rely on the order of the items returned from `GetConstructors()` and friends. I'd much prefer explicitly getting the parameterless constructor.

src/MonoMod.UnitTest/RuntimeDetour/ILCursorTest.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+            catch {
+                sb.Append("catch");
+            }
+            finally {
+                sb.Append("finally");
+            }
+            sb.Append("return");
+            return sb.ToString();
+        }
```
nike4613, line 181: This approach does worry me slightly, because it's dependent on how the compiler chooses to emit IL for this code. I don't expect it to actually be a problem, however I worry that over time these tests will start to fail over compiler version changes.

I'm fine with accepting this as-is, though, for now.

nike4613 (changes requested):

src/MonoMod.UnitTest/RuntimeDetour/ILCursorTest.cs
Chicken-Bones: The alternative is building the IL for this kind of test from scratch, which has its own drawbacks.

src/MonoMod.Utils/Cil/ILCursor.cs
Chicken-Bones: EH end markers point to the instruction following the end of the range, so they are not inclusive. When inserting an instruction after an EH range, the markers need to be updated to point to it, otherwise the inserted instruction would be inside the EH range.

Eg
```cs
try { ... } 
catch { ... 
    Prev
    // cursor is not here (before EH end)
}
// cursor is here (after EH end)
Next
```

nike4613: Ah, OK.

nike4613 (approved):

PR #183, opened 2024-06-02
Miscellaneous fixes to work properly on old Unity Mono

nike4613:


slxdy:
The timing of this is amazing. I was literally just looking for a solution for the RuntimeType issue lol.

Can we got a build of this soon? I'm working on net20 support for my mod loader, and otherwise I'd have to fall back to the older versions of MonoMod

nike4613:
Can always use the CI builds. That said, `net20`? I don't think there's *anything* these days that requires `net20`. (Even truly ancient Unity games target `net35`.)

slxdy:
Nope. There a lots of old mono games that use net20.
Also, the first commit in this PR is a fix specifically for net20

slxdy:
> Can always use the CI builds. That said, `net20`? I don't think there's _anything_ these days that requires `net20`. (Even truly ancient Unity games target `net35`.)

Do those builds also get uploaded to nuget?

nike4613:
> Nope. There a lots of old mono games that use net20. Also, the first commit in this PR is a fix specifically for net20

It's not, though. That's a fix that's present on old Mono, which is `net35`.

> Do those builds also get uploaded to nuget?

Not at the moment. I want to have them upload to GitHub Packages, but haven't set that up just yet.

slxdy:
Either way, I will just wait for the new nuget then. But thanks for merging this

PR #184, opened 2024-06-04
Remove Unsafe.As UB in {,ReadOnly}Span backports

Windows10CE:


src/MonoMod.ILHelpers/MonoMod.ILHelpers.il
```diff
@@ -31,33 +31,23 @@
...
-        ldloca str
-        conv.u
-        stloc ptrRef
-        ldloc ptrRef
+        
+        ldarg obj
+        stloc pin
+        
+        ldloca pin
         ldind.i
```
nike4613, line 61: IIRC I'd found that this doesn't work on the old Mono where we needed the leper. This case needs the `conv.u` and `stloc` to force a typecast, or it'll throw for invalid IL.

src/MonoMod.Backports/System/Memory,is_fx,lt_core_2.1,lt_std_2.1/Runtime/InteropServices/MemoryMarshal.Portable.cs
```diff
@@ -72,25 +72,13 @@ public static ReadOnlySpan<byte> AsBytes<T>(ReadOnlySpan<T> span)
...
         /// would have been stored. Such a reference can be used for pinning but must never be dereferenced.
         /// </summary>
-        public static ref T GetReference<T>(ReadOnlySpan<T> span)
-        {
-            if (span.Pinnable == null)
-                unsafe { return ref Unsafe.AsRef<T>(span.ByteOffset.ToPointer()); }
-            else
-                return ref Unsafe.AddByteOffset<T>(ref span.Pinnable.Data, span.ByteOffset);
-        }
+        public static ref T GetReference<T>(ReadOnlySpan<T> span) => ref Unsafe.AsRef(in span.GetPinnableReference());
```
nike4613, line 81: Do we want to make this return `ref readonly` as well? Is that a breaking change worth caring about?

nike4613 (changes requested):

src/MonoMod.ILHelpers/MonoMod.ILHelpers.il
```diff
@@ -34,24 +34,38 @@
...
+            !!T& finalRef
+#ifdef NET35
+            , !!T** refPtr
 #endif
         )
         
         ldarg obj
         stloc pin
         
+        // see https://nike4613.github.io/MonoMod/docs/RuntimeIssueNotes.html#fixed-on-strings-in-old-mono
```
nike4613: This document is a markdown file in this repo. Refer to that instead.

src/MonoMod.Backports/System/Memory,is_fx,lt_core_2.1,lt_std_2.1/Runtime/InteropServices/MemoryMarshal.Portable.cs
Windows10CE: The signature was already like that, i'd say leave it https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.getreference?view=net-8.0#system-runtime-interopservices-memorymarshal-getreference-1(system-readonlyspan((-0)))

src/MonoMod.ILHelpers/MonoMod.ILHelpers.il
```diff
@@ -34,24 +34,38 @@
...
+#else
         .locals (
-            object pinned pin
-#ifndef netcoreapp
-            , !!T& ref
+            object pinned pin,
+            !!T& finalRef
+#ifdef NET35
+            , !!T** refPtr
 #endif
```
nike4613: I'm not a huge fan of gating this behavior behind NET35. I'm not actually sure when the issue is fixed; it's possible its not fixed until sometime after Mono transitioned to .NET Framework 4.x; I'd like to at least double check that, or otherwise always do the extra casting.

src/MonoMod.UnitTest/Temp.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class Temp : TestBase
+    {
+        public Temp(ITestOutputHelper helper) : base(helper) { }
+
+        [Fact]
+        public void MyThing()
+        {
+            Assert.Equal("abc", "abc".AsSpan().ToArray());
+        }
+    }
```
nike4613: Can we have an actual name for this test?

nike4613 (changes requested):

src/MonoMod.Backports/System/Memory,is_fx,lt_core_2.1,lt_std_2.1/Runtime/InteropServices/MemoryMarshal.Portable.cs
nike4613: Fair enough.

src/MonoMod.UnitTest/Temp.cs
Windows10CE: I just deleted it instead, was just using it for debugging, but i can add some tests for general span correctness if you'd like

nike4613: Probably not a bad idea. At some point, we should also add some mini test framework that we can run on .NET 3.5...

nike4613 (approved):

PR #192, opened 2024-09-09
.NET 9 Support

nike4613:
In addition to adding an updated JIT GUID and vtable offsets, this adds `IControlFlowGuard`, as .NET 9 now builds with CFG enabled.


nike4613:
This is already merged, GitHub is just having a moment.

PR #197, opened 2024-11-16
Add an ABI fixup for generic context

### LINKED ISSUE #185: Arguments passed to a hook from a generic method are shifted by 1
### Description

Hi! I noticed a strange off-by-one error while trying to hook into a generic method. I understand that this library doesn't fully support generics due to the numerous complications they present. However, this problem does seem like a simple stray `-1` somewhere it shouldn't be, so maybe it can be easily fixed.

### Example

Please consider the following MRE:

```cs
var addLast = typeof(Foo).GetMethod(nameof(Foo.AddLast))!.MakeGenericMethod(typeof(string));
var addLastHook = typeof(FooHook).GetMethod(nameof(FooHook.AddLast))!.MakeGenericMethod(typeof(string));

var hook = new Hook(addLast, addLastHook);

new Foo().AddLast(new(), "Hello world");

// ...

public class Foo
{
    public void AddLast<T>(List<T> list, T item)
    {
        list.Remove(item);
        list.Add(item);
    }
}

public static class FooHook
{
    public static void AddLast<T>(Action<Foo, List<T>, T> addLast, Foo it, List<T> list, T item)
    {
        // `addLast` is fine and works properly.
        //
        // `it` is also fine and indeed appears as a `Foo` instance.
        //
        // `item` is actually `List<T>`, not `T`, while
        // `list` is not even an object.
        //
        // Any attempt to do literally anything with `list`
        // will result in a `NullReferenceException` being thrown.
        // Even operations like `Console.WriteLine(list is List<string>)`!
        addLast(it, list, item);
    }
}
```

In this case, `item` receives a value that was supposed to be passed as `list`, while `list` seemingly hasn't received anything at all and represents a chunk of uninitialized memory *(it is not a valid object, and sometimes I get an `AccessViolationException` for trying to inspect the memory behind it)*.

This pattern continues if we keep adding parameters:
- The callback *(if any)* and the instance reference *(if any)* are set properly.
- The next argument is totally borked.
- The next argument after that receives the value of the previous one.
- The next argument after that receives the value of the previous one, too.
- ...
- The next argument after that receives the value of the previous one, too. 
- There's no way to access the last argument passed to the original method.

I hope you get the idea. Do you have any thoughts on where the problem could stem from?

Kira-NT:
### Description

Currently, if you attempt to inject into a generic method while using .NET, the injection itself works, but the hook often receives incorrect arguments. This issue arises because .NET passes a generic context to some generic methods in the form of `MethodDesc`/`MethodTable`/`TypeHandle`, which shifts all user arguments by one position.

`MonoMod` already supports the concept of "ABI fixups," so I resolved the issue by implementing a new ABI fixup rule for these scenarios. This rule disregards the provided context, similarly to how it's currently done in the Mono realm. In Mono, the `r10` register is used instead of the hidden argument for this purpose, so it doesn't create such a problem, and we can implicitly ignore its existence.

While this fix does not alter the official stance that "MonoMod doesn't support generics," it allows generics to function in .NET under conditions where they already somewhat work in Mono. Thus, this can be considered a parity feature :)

### Examples

Starting with .NET 5, all the following methods are now injectable:

```csharp
public class Foo
{
    public static void StaticMethod<T>(T t);

    public void InstanceMethod<T>(T t);
}

public class Bar<T>
{
    public static void StaticMethod(T t);

    public void InstanceMethod(T t);

    public static void StaticMethod2<U>(T t, U u);

    public void InstanceMethod2<U>(T t, U u);
}

public struct Baz
{
    public static void StaticMethod<T>(T t);

    public void InstanceMethod<T>(T t);
}

public struct Qux<T>
{
    public static void StaticMethod(T t);

    public void InstanceMethod(T t);

    public static void StaticMethod2<U>(T t, U u);

    public void InstanceMethod2<U>(T t, U u);
}
```

However, since this PR does not aim to address deeper issues, some existing problems still persist. For instance, certain injections fail to work on Windows *(though they function perfectly on Linux and macOS)* because MonoMod silently fails to apply them. Here are some scenarios I have identified:

```csharp
public void InstanceVoid<T>(List<T> list, T item, ref bool injected);

public string InstanceRegister<T>(List<T> list, T item, ref bool injected);

public LargeStruct InstanceByRef<T>(List<T> list, T item, ref bool injected);

public static LargeStruct StaticByRef<T>(List<T> list, T item, ref bool injected);
```

Thankfully, most common use cases "just work" now. And honestly, it's super useful when they do!

### Notes

Fixes #185

### DIFF
--- a/src/MonoMod.Core/Platforms/PlatformTriple.cs
+++ b/src/MonoMod.Core/Platforms/PlatformTriple.cs
@@ -661,8 +661,35 @@ namespace MonoMod.Core.Platforms
             var hasThis = !fromInfo.IsStatic;
             var requiresReturnBufferFixup = hasThis && toInfo.IsStatic && hasReturnBuffer;
 
+            // Whenever we detour a call from a generic method, depending on the ABI, we may
+            // receive a generic context as an argument, which a callee never needs, at least
+            // in the form of an explicitly passed argument, even if it is also a generic method.
+            // The reason here is simple: we do NOT allow specifying a generic method definition
+            // as a detour target (while that would make sense from a user's perspective, it
+            // requires much more work to function properly), thus, `to` can only be
+            // a constructed generic method with its generic context baked in.
+            // This leads to a pretty bad discrepancy:
+            //
+            // Caller's argument order: [ThisPointer, ReturnBuffer, GenericContext, UserArguments]
+            // Callee's argument order: [ThisPointer, ReturnBuffer, UserArguments]
+            //
+            // Because of this, the callee will receive a pointer to the generic context information
+            // instead of its first argument, causing all subsequent arguments to be shifted by one,
+            // resulting in all types of phantasmagorical problems that will immediately blow up in
+            // the user's face. Moreover, to add insult to injury, there is no way to access
+            // the last argument, although this is the least of our concerns here.
+            //
+            // Note that we only need this fixup if the current ABI explicitly defines the generic
+            // context as an argument. For example, Mono on Windows/Wine x86_64 stores a pointer to
+            // the generic context struct in r10, so generic injections somewhat work there already.
+            //
+            // Consequently, there is no need for a fixup for ABIs that declare `GenericContext`
+            // **after** `UserArguments`. However, as far as I know, this is only relevant for
+            // RyuJITx86, so there is no value in optimizing for that scenario.
+            var requiresGenericContextFixup = HasGenericContext(Abi) && RequiresGenericContext(fromInfo);
+
             // Check if a fixup is needed for the current scenario.
-            if (!requiresReturnBufferFixup)
+            if (!requiresReturnBufferFixup && !requiresGenericContextFixup)
             {
                 // `from` and `to` are considered compatible at this point.
                 // Thus, no ABI fixups are needed, return `to` as is.
@@ -692,14 +719,45 @@ namespace MonoMod.Core.Platforms
                         argumentTypes.Add(returnBufferType);
                         break;
 
+                    case SpecialArgumentKind.GenericContext when requiresGenericContextFixup:
+                        // Currently, we do the bare minimum: we acknowledge that
+                        // the generic context exists. That's all. After that,
+                        // we simply throw it out of the window and rely on
+                        // the generic context (if any) baked into in the callee.
+                        //
+                        // While this does work fine, it introduces funny little
+                        // holes in the type-safety premise of .NET:
+                        //
+                        // // This may return `false`!
+                        // bool Hook<T>(T it)
+                        //     => typeof(T).IsAssignableFrom(it.GetType());
+                        //
+                        // This behavior is caused by the fact that constructed
+                        // generics for reference types are Javad into a single
+                        // definition at runtime (and rightfully so).
+                        // So, even if you try to detour a specific generic
+                        // implementation using something like
+                        // `to.MakeGenericMethod(typeof(string))`,
+                        // your hook will receive calls for all
+                        // reference type-based implementations out there.
+                        // However, since we do not patch the generic context
+                        // of the provided hook, it remains unchanged,
+                        // causing `typeof(T)` to defy users' expectations.
+                        //
+                        // So, TODO: patch the generic context of the detour target
+                        // and/or introduce a call filter based on the current context.
+
+                        // The generic context is passed as a pointer to a struct that
+                        // contains all the needed (?) information.
+                        // We can treat it as a simple `IntPtr`.
+                        argumentTypes.Add(typeof(nint));
+                        break;
+
                     case SpecialArgumentKind.UserArguments:
                         userArgumentsOffset = argumentTypes.Count;
                         argumentTypes.AddRange(parameters.Select(static p => p.ParameterType));
                         break;
                 }
-
-                // TODO: somehow handle generic context parameters
-                // or more likely, just ignore it for this and do generics elsewhere
             }
 
             Helpers.DAssert(thisPos >= 0 || !hasThis);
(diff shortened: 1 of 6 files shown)


src/MonoMod.Core/Platforms/PlatformTriple.cs
```diff
@@ -626,106 +626,234 @@ public MethodBase GetRealDetourTarget(MethodBase from, MethodBase to)
...
+                        // implementation using something like
+                        // `to.MakeGenericMethod(typeof(string))`,
+                        // your hook will receive calls for all
+                        // reference type-based implementations out there.
+                        // However, since we do not patch the generic context
+                        // of the provided hook, it remains unchanged,
+                        // causing `typeof(T)` to defy users' expectations.
+                        //
+                        // So, TODO: patch the generic context of the detour target
+                        // and/or introduce a call filter based on the current context.
```
nike4613, line 748: This change in behavior (towards silently not doing what the user wants) drives me to want to also block generics in the RuntimeDetour API (though I'm OK with leaving it open from the Core API).

src/MonoMod.Core/Platforms/PlatformTriple.cs
```diff
@@ -626,106 +626,234 @@ public MethodBase GetRealDetourTarget(MethodBase from, MethodBase to)
...
+            var mayNeedHiddenArg = method.IsGenericMethod
+                || method.IsStatic
+                || declaringType.IsValueType
+                || declaringType.IsInterface && !method.IsAbstract;
+            if (!mayNeedHiddenArg)
+                return false;
+
+            // Finally, check whether the type on which the method is defined may be shared.
+            var typeGenericArguments = declaringType.IsGenericType ? declaringType.GetGenericArguments() : Type.EmptyTypes;
+            return typeGenericArguments.Any(static x => !x.IsValueType);
```
nike4613: All of this logic is runtime-specific, and so should call some runtime-specific helper to make this determination. The rules implemented here are also wrong: it mishandles the case of `M<SomeStruct<string>>()` *which is still shared*, even though the type argument is a value-type. (I'll point you over [here](https://github.com/nike4613/MonoMod/tree/coreclr-generics) and [here](https://github.com/MonoMod/MonoMod.Common/tree/15d444c207a49232318f423392177f0dbea82548/RuntimeDetour/Platforms/Generic) for something to reference-- I'm pretty sure I got the logic fully correct there.)

nike4613 (changes requested):
I'm interested in merging this, if only so that we can have this part ready when we *do* actually get around to generics. It does need some work though, and (as I mentioned in one of the other comments) I would want to block generics in RuntimeDetour with this, precisely because the effects of trying to patch generics are less catastrophic.

src/MonoMod.Core/Platforms/PlatformTriple.cs
Kira-NT: This makes sense, especially to avoid new issues like, *"Generic injections don't work the way I want!"* Where would you prefer to place this roadblock? I believe `Hook.PrepareRealTarget` could be a good candidate for that, as it already handles signature checks and similar stuff.

src/MonoMod.Core/Platforms/PlatformTriple.cs
Kira-NT: > All of this logic is runtime-specific

Yup, I initially wanted to put it in `IRuntime`, but that would be a breaking change since it impacts a public API. So, instead, I decided to place the method here to avoid a more significant change, as it realistically only affects CoreCLR at the moment :)

If you're okay with `IRuntime.RequiresGenericContext(MethodBase)` *(or something like that)*, I'll be happy to fix it and move the method to `Core[...]Runtime` while creating stubs for the other runtimes.  

> it mishandles the case of `M<SomeStruct<string>>()`

Oh, right - great catch! I basically followed the same CLR logic you referenced in the comments of your code but made the mistake of trivializing shared generic types to just non-value types, while somehow forgetting in the moment that value types can also be generic and, thus, shared :D

src/MonoMod.Core/Platforms/PlatformTriple.cs
nike4613: That's probably where I'd put it, yes.

src/MonoMod.Core/Platforms/PlatformTriple.cs
nike4613: The way I'm treating Core versioning is that any non-*consumer* breaking change is permitted across minor version bumps, and that certain consumer-breaking changes are permitted as long as they're not at the main entrypoint (i.e. DetourFactory). If we really don't like adding a method to `IRuntime`, then we can also add another interface to the implementers of `IRuntime`, like I did for CET, but I'd prefer just changing `IRuntime`.

src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
```diff
@@ -161,6 +162,62 @@ public virtual RuntimeMethodHandle GetMethodHandle(MethodBase method)
...
+        private static bool IsGenericSharedType(Type type)
+        {
+            // Primitives are never shared.
+            // Value types are shared only if at least one of their nested generic arguments is shared.
+            // All other types (i.e., reference types) are always shared.
+            return (
+                !type.IsPrimitive &&
+                (!type.IsValueType || type.IsGenericType && type.GetGenericArguments().Any(IsGenericSharedType))
+            );
+        }
```
nike4613, line 219: It might be worth caching this somewhere, possibly in a `CWT<Type, StrongBox<bool>>`? Esp. seeing as it can recurse arbitrarily deeply. Not sure how important that is though.

nike4613 (approved):

src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
Kira-NT: While it *theoretically* can recurse quite deeply, I don't think it's worth optimizing for this scenario in practice. For this method to become substantially computationally expensive, we would need to encounter a deeply nested generic value type in the form of `Struct0<Struct1<Struct2<Struct3<...>, ...>, ...>, ...>` with no reference-type generic arguments to break the loop. And even then, how deep would it really need to be to cause a problem? Even a hundred levels of nesting would be processed in an instant; and in all the years I've worked with .NET, the worst atrocities I've encountered haven't gone beyond "just" a few dozens of total generic arguments.

This is such an insanely rare *(if not outright unrealistic)* scenario that, to be quite frankly honest with you, if someone does create types like this, they deserve to face the consequences of their actions! :D

Given this, I feel like polluting the runtime with a permanent cache for something we might not even encounter a second time would be a significantly worse tradeoff than simply computing the result of this relatively cheap operation as is.

nike4613: I have written such code, though in general I do agree.

PR #202, opened 2024-12-09
One-character null reference fix

connorivy:
When trying to build an ILHook, I get an exception during the 'DynamicMethodDefinition.Generate' method. It turns out that reflection helpers.GetRuntimeHashedFullName() is being passed an  AssemblyNameReference object with a hash algorithm, yet the hash property is null.

Looking at the definition of AssemblyNameReference in Mono.Cecil, the hash object is not necessarily set in the constructor, so it seems reasonable to assume that it may be null.

After I made this one character fix, my ILHook started to work exactly as expected.


nike4613:
Can you add a test which would otherwise trigger this error?

PR #207, opened 2024-12-30
Fix a bug when executing in an isolated load context.

EvilMask:
[`MonoMod.Utils.Cil.ILGeneratorShim.GetProxy()`](https://github.com/MonoMod/MonoMod/blob/26a0a4d22fcd233dd22db4fd034cb3ac41420fd4/src/MonoMod.Utils/Cil/ILGeneratorShim.Proxy.cs#L25) will throw exception:
```
System.ArgumentException: GenericArguments[0], 'MonoMod.Utils.Cil.CecilILGenerator', on 'MonoMod.Utils.Cil.ILGeneratorProxy[TTarget]' violates the constraint of type 'TTarget'.
```
when MonoMod is executing in an isolated load context. (The situation as I mentioned here: pardeike/Harmony#642)


EvilMask:
Seems not to be a general solution, close it for now.

PR #208, opened 2024-12-30
MonoModder support multi mods patch same method

MikiraSora:
As the documentation said, currently MonoModder does not support multiple mods patching the same method.

The current situation is that if multiple mods patch the same method, the patched methods are replaced, not merged together. This is unacceptable for multiple mod developers, as they don't want their mod implementations to be dropped or disabled by other mods.

Therefore, I will try to add new conditions and implementations to MonoModder. The conditions and replacement process are written in code comments.

However, this feature is turned off by default for current compatibility, so if you need this feature, please set `CombineSameMethodMultiModPatches`(or env var `MONOMOD_COMBINE_MORE_MODS_PATCH_SAME_METHOD=1`) to true.

Also added some related unit tests


nike4613 (changes requested):
I very much dislike having opaque binaries checked in to the source tree, *especially* without the source which produces them being checked in. My preference would be building the assemblies in the tests using Cecil instead.

PR #212, opened 2025-02-04
Enable `noexecstack` linker arg for x86_64 linux exeption helper

### LINKED ISSUE #198: Unable to Load Temporary Native Library due to executable stack Restriction
When using MonoMod.Core via HarmonyLib, an error occurs during the dynamic generation and loading of the native library (/tmp/mm-exhelper.so.XXXXXX) under systems with stricter security policies. The issue arises because the temporary native library requires an executable stack, which is prohibited in certain environments. This results in the following exception:
`System.DllNotFoundException: Unable to load shared library '/tmp/mm-exhelper.so.wGm60g' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: /tmp/mm-exhelper.so.wGm60g: cannot enable executable stack as shared object requires: Permission denied`

This happens because the dynamically created library requires an executable stack, which is restricted by the host system's security settings. This issue is particularly problematic in environments where:
- Policies such as noexec or SELinux/AppArmor block such operations.
- Executable stacks are explicitly disabled by the operating system.

covers1624:
Enables the `-z noexecstack` linker argument.

@Chicken-Bones Has kindly tested this change with multiple tML users who were suffering this issue, and reported that the issue was resolved. It appears a recent ArchLinux update has sparked an influx of this issue for them.

Fixes #198


nike4613:
GitHub changed the software in the runners it looks like. It seems like the change is good though.

Issue #223, opened 2025-04-11
Import different assemblies with the same name breaks MonoMod Reflection Importer/Resolver


wuke32767:
### Description
see title.

Here is what I've found:
https://github.com/MonoMod/MonoMod/blob/bc177577c97ff9d52074f258f2d0735cc2416603/src/MonoMod.Utils/MMReflectionImporter.cs#L121
Imported assembly name is *cached*. If a different assembliy with the same name was imported, it would modify the previous hash.
### Example
```cs
class MyClass
{
    public static void Main(string[] args)
    {
        AssemblyLoadContext a = new("a");
        var x = a.LoadFromAssemblyPath(typeof(MyClass).Assembly.Location);
        DynamicMethodDefinition dmd = new("", typeof(object), []);
        ILContext il = new(dmd.Definition);
        var c = il.Import(x.EntryPoint);
        var d = il.Import(typeof(MyClass).Assembly.EntryPoint);
        var e = c.ResolveReflection() == d.ResolveReflection();
        if (e)
        {
            throw new Exception();
        }
    }
}
```
Intended: nothing happen (`c` and `d` corresponding to the different methodinfo)
Actual: crash

nike4613:
MonoMod doesn't work well with ALCs in general. The linked PR has tests for this.

PR #224, opened 2025-04-11
Fix alc import things

wuke32767:
Finally i solved the #223.

Based on #178.

One line in the test case was removed. It will hook an alc method, so that it will be trapped in the DetourManager forever, can't be collected.

it's not clean, and should be improved.


wuke32767:
anyway i have checked it for (really) many times, there is no evidence to suggest that new cacheKey did not include everything.

time to reopen.

src/MonoMod.UnitTest/RuntimeDetour/AssemblyLoadContextHookTest.cs
```diff
@@ -197,6 +200,9 @@ public static void TestAssemblyLoadContextHookLoaded(object loader, int id1, int
                 method.Invoke(null, new object[] { null, -1, -1 });
             }
             verify.Invoke(loader, argsSet);
+
+            TestILHookReferences(loader.GetType(), loader.GetType());
+            //TestILHookReferences(loader.GetType(), typeof(AssemblyLoadContextHookTest));
```
nike4613: Is this correct? Doesn't this end up being the "trivial" case? (`!alc`)

src/MonoMod.Utils/ReflectionHelper.cs
```diff
@@ -103,6 +103,27 @@ public static Assembly Load(Stream stream)
...
+                    if (hash[i] != AssemblyHashPrefix[i])
+                        return false;
+                var rest = BitConverter.GetBytes(asm.GetHashCode());
+
+                for (var i = 0; i < 4; i++)
+                    if (hash[AssemblyHashPrefix.Length + i] != rest[i])
+                        return false;
+                return true;
+            }
+            return truewhennohash;
```
nike4613: This should probably be named `defaultIfNoHash` or similar instead

src/MonoMod.UnitTest/RuntimeDetour/AssemblyLoadContextHookTest.cs
wuke32767: (alc)AssemblyLoadContextHookTest.ILHookTarget will be hooked, so (non-alc)DetourManager will reference it, prevents it from being collected.
see https://github.com/wuke32767/MonoMod/actions/runs/16742036668 for test result

nike4613 (approved):

PR #225, opened 2025-04-21
x86 detour of instance method with params returning 2 or 4 byte structs crashes

pardeike:


pardeike:
This needs a better fix than simply checking against return type sizes of 2 and 4 but all tests pass!

pardeike:
This needs a better fix than simply checking against return type sizes of 2 and 4 but all tests pass!

pardeike:
This needs a better fix than simply checking against return type sizes of 2 and 4 but all tests pass!

src/MonoMod.Core/Platforms/PlatformTriple.cs
```diff
@@ -10,6 +10,7 @@
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
```
nike4613: Lets not modify files that don't need to be :)

src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
```diff
@@ -68,8 +68,12 @@ TypeCode.Int32 or TypeCode.UInt32 ||
...
             // all others are passed on stack, or in a return buffer
             if (isReturn)
-            {
-                return TypeClassification.ByReference;
+            {
+                var typeSize = type.GetManagedSize();
+                if (typeSize is 1 or 2 or 4)
+                    return TypeClassification.InRegister;
+                else
+                    return TypeClassification.ByReference;
```
nike4613, line 76: I can't say I'm super happy with this, as is. I think at some point we should go and inspect the ABI logic for x86 and redo it from that. That's for another time though I think.

src/MonoMod.UnitTest/Core/Return2ByteStructIn32Bit.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MonoMod.UnitTest.Core
+{
+    [Obsolete("This test is replaced by DetouringMethodsReturningStructs")]
+    public class Return2ByteStructIn32Bit : TestBase
```
nike4613: If this is replaced, we don't need it at all, no?

src/MonoMod.UnitTest/Core/Return4ByteStructIn32Bit.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MonoMod.UnitTest.Core
+{
+    [Obsolete("This test is replaced by DetouringMethodsReturningStructs")]
+    public class Return4ByteStructIn32Bit : TestBase
```
nike4613: See above

test.runsettings
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+  <RunConfiguration>
+    <TargetPlatform>x86</TargetPlatform>
+  </RunConfiguration>
+</RunSettings>
```
nike4613: This should probably be gitignored and not checked in.

nike4613 (changes requested):

src/MonoMod.Core/Platforms/PlatformTriple.cs
pardeike: left over

src/MonoMod.UnitTest/Core/Return2ByteStructIn32Bit.cs
pardeike: yes

src/MonoMod.UnitTest/Core/Return4ByteStructIn32Bit.cs
pardeike: can go too

test.runsettings
pardeike: I'll add *.runsettings to .gitignore

src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
pardeike: Could this be the place @nike4613 ?
https://github.com/dotnet/runtime/blob/29acb2fd71a363053f1fbe67dc4bc06c555f7b94/src/coreclr/jit/compiler.cpp#L906

nike4613: It looks like it, yes.

nike4613 (approved):

PR #240, opened 2025-06-11
Fix calli

### LINKED ISSUE #229: Calli instruction is broken
### Description
Calli is broken. Only DynamicMethod backend is broken.

### Example
```cs
using Mono.Cecil.Cil;
using MonoMod;
using MonoMod.Core.Platforms;
using MonoMod.Utils;

DynamicMethodDefinition dmd = new("a", null, [typeof(nint),typeof(valid)]);
var il = dmd.GetILProcessor();
var c = new Mono.Cecil.CallSite(dmd.Module.ImportReference(typeof(void)));
c.Parameters.Add(new(dmd.Module.ImportReference(typeof(valid))));
il.Emit(OpCodes.Ldarg_1);
il.Emit(OpCodes.Ldarg_0);
il.Emit(OpCodes.Calli, c);
il.Emit(OpCodes.Ret);
//Switches.SetSwitchValue(Switches.DMDType, "cecil");
PlatformTriple.Current.Compile(dmd.Generate());
public struct valid { }
```
results:
```
Unhandled exception. System.TypeLoadException: Could not load type 'MonoMod.Switches' from assembly 'MonoMod.Utils, Version=25.0.8.0, Culture=neutral, PublicKeyToken=null' due to value type mismatch.
   at System.Runtime.CompilerServices.RuntimeHelpers.CompileMethod(RuntimeMethodHandleInternal method)
   at invoke RuntimeHelpers.CompileMethod(RuntimeMethodHandle )
   at MonoMod.Core.Platforms.Runtimes.FxCoreBaseRuntime.TryInvokeBclCompileMethod(RuntimeMethodHandle handle)
   at MonoMod.Core.Platforms.Runtimes.FxCoreBaseRuntime.GetMethodHandle(MethodBase method)
   at MonoMod.Core.Platforms.Runtimes.FxCoreBaseRuntime.Compile(MethodBase method)
   at MonoMod.Core.Platforms.PlatformTriple.Compile(MethodBase method)
   at Program.<Main>$(String[] args)
```

wuke32767:
closes #229.
i don't know if it actually works, but all the test cases was passed.

also contains new test cases. it seems that i can't get function pointer of a generic method on mono, so i wrote a simpler test case for mono.

### DIFF
--- a/src/MonoMod.UnitTest/CalliTest.cs
+++ b/src/MonoMod.UnitTest/CalliTest.cs
@@ -1,0 +1,131 @@
+﻿using Mono.Cecil.Cil;
+using MonoMod.Core.Platforms;
+using MonoMod.Utils;
+using System;
+using Xunit;
+using Xunit.Abstractions;
+
+
+namespace MonoMod.UnitTest
+{
+    public unsafe sealed class CalliTests : TestBase
+    {
+        public CalliTests(ITestOutputHelper helper) : base(helper)
+        {
+        }
+
+        [Fact]
+        public void CalliS()
+        {
+            if (PlatformDetection.Runtime is RuntimeKind.Mono)
+            {
+                LightCalli<random_struct>();
+            }
+            else
+            {
+                Calli<random_struct>();
+            }
+        }
+        [Fact]
+        public void CalliO()
+        {
+            if (PlatformDetection.Runtime is RuntimeKind.Mono)
+            {
+                LightCalli<random_class>();
+            }
+            else
+            {
+                Calli<random_class>();
+            }
+        }
+        delegate T Helper<T>(T o, ref T r, T[] a, T* p) where T : new();
+        public static unsafe T Ret<T>(T o, ref T r, T[] a, T* p) where T : new()
+        {
+            Assert.Equal(typeof(T), o.GetType());
+            Assert.Equal(typeof(T[]), a.GetType());
+            Assert.Equal(r, *p);
+            r = new T();
+            Assert.Equal(r, *p);
+            return o;
+        }
+        unsafe void Calli<T>() where T : new()
+        {
+            T obj = new();
+            Ret(obj, ref obj, [obj], (T*)Unsafe.AsPointer(ref obj));
+
+            Type type = typeof(T);
+            var method = ((Helper<T>)Ret).Method;
+            var i = method.MethodHandle.GetFunctionPointer();
+            using DynamicMethodDefinition dmd = new("a", null, [typeof(nint), type]);
+            var il = dmd.GetILProcessor();
+            var c = new Mono.Cecil.CallSite(dmd.Module.ImportReference(type));
+            c.Parameters.Add(new(dmd.Module.ImportReference(type)));
+            c.Parameters.Add(new(dmd.Module.ImportReference(type.MakeByRefType())));
+            c.Parameters.Add(new(dmd.Module.ImportReference(type.MakeArrayType())));
+            c.Parameters.Add(new(dmd.Module.ImportReference(type.MakePointerType())));
+            il.Emit(OpCodes.Ldarg_1);
+            il.Emit(OpCodes.Ldarga, 1);
+            il.Emit(OpCodes.Ldc_I4_0);
+            il.Emit(OpCodes.Newarr, type);
+            il.Emit(OpCodes.Ldarga, 1);
+            il.Emit(OpCodes.Ldarg_0);
+            il.Emit(OpCodes.Calli, c);
+            il.Emit(OpCodes.Pop);
+            il.Emit(OpCodes.Ret);
+            //Switches.SetSwitchValue(Switches.DMDType, "cecil");
+            var ret = dmd.Generate();
+            PlatformTriple.Current.Compile(ret);
+            ret.Invoke(null, [i, Activator.CreateInstance(type, [32])]);
+        }
+        static void Light(object u)
+        {
+            throw new NotImplementedException("MonoMod.Test.Intended" + u.ToString());
+        }
+        unsafe void LightCalli<T>() where T : new()
+        {
+
+            Type type = typeof(T);
+            var method = ((Delegate)Light).Method;
+            var i = PlatformTriple.Current.GetNativeMethodBody(method);
+            using DynamicMethodDefinition dmd = new("a", null, [typeof(nint), type]);
+            var il = dmd.GetILProcessor();
+            var c = new Mono.Cecil.CallSite(dmd.Module.TypeSystem.Void);
+            c.Parameters.Add(new(dmd.Module.ImportReference(type.MakeArrayType())));
+            il.Emit(OpCodes.Ldc_I4_0);
+            il.Emit(OpCodes.Newarr, type);
+            il.Emit(OpCodes.Ldarg_0);
+            il.Emit(OpCodes.Calli, c);
+            il.Emit(OpCodes.Ret);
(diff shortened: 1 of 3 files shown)


src/MonoMod.Utils/DMDGenerators/DMDEmit.EmitCallSite.cs
```diff
@@ -4,6 +4,7 @@
 using System.Reflection;
 using System.Reflection.Emit;
 using System.Runtime.InteropServices;
+using static MonoMod.Utils.Interop.Windows;
```
nike4613: Remove pls

src/MonoMod.Utils/DMDGenerators/DMDEmit.EmitCallSite.cs
```diff
@@ -462,6 +463,12 @@ void AddOneArgTypeHelperWorker(Type clsArgument, bool lastWasGenericInst)
                         {
                             AddElementType(type);
                         }
+                        // https://github.com/dotnet/runtime/blob/4e0dc086de5952b87e0fdac43690edb5acc4bd6b/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/SignatureHelper.cs#L449
+                        // btw module is always null
+                        else if (true || tokenCreator/*signatureHelper.module*/ == null)
+                        {
+                            InternalAddRuntimeType(clsArgument);
+                        }
```
nike4613: Is this the case on .NET Framework as well? Can we add a link to referencesource for the impl?

src/MonoMod.UnitTest/CalliTest.cs
```diff
@@ -0,0 +1,133 @@
...
+        delegate T Helper<T>(T o, ref T r, T[] a, T* p) where T : new();
+        public static unsafe T Ret<T>(T o, ref T r, T[] a, T* p) where T : new()
+        {
+            Assert.Equal(typeof(T), o.GetType());
+            Assert.Equal(typeof(T[]), a.GetType());
+            Assert.Equal(r, *p);
+            r = new T();
+            Assert.Equal(r, *p);
+            return o;
+        }
```
nike4613: What's going on here? What purpose does this serve?

src/MonoMod.UnitTest/CalliTest.cs
```diff
@@ -0,0 +1,133 @@
...
+        unsafe void Calli<T>() where T : new()
+        {
+            T obj = new();
+            Ret(obj, ref obj, [obj], (T*)Unsafe.AsPointer(ref obj));
```
nike4613: `Unsafe.AsPointer` shouldn't be needed here. You should be able to just do `&obj`, and that would be preferable.

src/MonoMod.UnitTest/CalliTest.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            }
+        }
+        public class random_class(int v)
+        {
+            int u = v;
+            public random_class() : this(0)
+            {
+            }
+        }
```
nike4613: Can we keep these names as PascalCase?

src/MonoMod.UnitTest/CalliTest.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+        }
+
+        [Fact]
+        public void CalliS()
+        {
+            if (PlatformDetection.Runtime is RuntimeKind.Mono)
+            {
+                LightCalli<random_struct>();
+            }
```
nike4613: Why do we need to do something different on Mono, and what are the differences in implementation?

nike4613 (changes requested):

nike4613:
Also, do you think you can merge in (and fix) https://github.com/MonoMod/MonoMod/pull/235 just so we have the exact repro case you posted in?

wuke32767:
#229 is the minimum reproduction step. to be more robust, i wrote a complex test case.

src/MonoMod.UnitTest/CalliTest.cs
wuke32767: test if it receives valid arguments

nike4613: Can you add a comment to that effect? It's not obvious why any of these asserts would fail under normal circumstances

nike4613:
> #229 is the minimum reproduction step. to be more robust, i wrote a complex test case.

I'd like to have both, especially since that one is simpler to read. More tests don't hurt.

src/MonoMod.UnitTest/CalliTest.cs
wuke32767: they should not fail under normal circumstances; if it fails, it mainly due to internal runtime error: in this case these assertions seem to be not needed.

but i would still verify these parameters, because i don't trust runtime, especially when it depends on private code.

src/MonoMod.UnitTest/Github/Issue229_CalliInstructionTest.cs
```diff
@@ -0,0 +1,30 @@
+using MonoMod.Utils;
+using Xunit;
+using Mono.Cecil.Cil;
+using MonoMod.Core.Platforms;
+using System;
+
+namespace MonoMod.UnitTest.Github
+{
+    public class Issue229_CalliInstructionTest
+    {
```
nike4613: Can you make this inherit the test base? The robot messed that up :)

nike4613 (approved):
Other than the one little thing, LGTM

src/MonoMod.Utils/DMDGenerators/DMDEmit.EmitCallSite.cs
```diff
@@ -462,6 +462,15 @@ void AddOneArgTypeHelperWorker(Type clsArgument, bool lastWasGenericInst)
                         {
                             AddElementType(type);
                         }
+                        // https://github.com/dotnet/runtime/blob/4e0dc086de5952b87e0fdac43690edb5acc4bd6b/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/SignatureHelper.cs#L449
+                        // https://referencesource.microsoft.com/#mscorlib/system/reflection/emit/signaturehelper.cs,491
+                        // btw dynamic method il generator emitcalli never gives it a module, so it is always null
+                        // https://github.com/dotnet/runtime/blob/4e0dc086de5952b87e0fdac43690edb5acc4bd6b/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/DynamicILGenerator.cs#L450
+                        // https://referencesource.microsoft.com/#mscorlib/system/reflection/emit/dynamicilgenerator.cs,538
+                        else if (true || tokenCreator/*signatureHelper.module*/ == null)
```
Chicken-Bones, line 470: `true ||`?

nike4613: Meh, I don't particularly care. I'm pretty sure Roslyn winds up removing it entirely anyway.

wuke32767: i want to keep the original code, but `signatureHelper.module` seems to correspond to `tokenCreator`, which is not null.

Chicken-Bones: Mostly just that it makes the block/case below it unreachable and I want to make sure that's intentional

Chicken-Bones: Normally we'd remove unreachable code

PR #241, opened 2025-06-27
Linux and MacOS arm64 support

### LINKED ISSUE #50: RuntimeDetour selftest failing on armv7, Android 4.X / 5.X
Thanks for your great job.I integrated the project to my Stardew Valley Mod Loader, and received some crash report on Android 5.1.1 and earlier version.
Here is the error log:
DetourRuntimeILPlatform._SelftestGetStruct (System.IntPtr x, System.IntPtr y)
System.Exception: This method should've been detoured!
screenshot

### LINKED ISSUE #90: Supporting Apple Silicon
I spent some time debugging over the weekend and figured out the necessary changes for supporting Apple Silicon. I can try to restructure the code a bit and send in a pull request, but I just wanted to write down the things that I encountered.

## W^X
Apple requires all pages to be W^X, and requires pages to be mapped in with the `MAP_JIT` flag.
1. When applying detours, the relevant pages are already mapped in with the `MAP_JIT` flag. We just need to do the following:
    - Call `pthread_jit_write_protect_np(false)` (to enable write and disable execute)
    - Make the changes
    - Call `pthread_jit_write_protect_np(true)` (to disable write and enable execute)
1. When creating the trampoline for the JIT hook, we need to `mmap` in memory with the `MAP_JIT` flag instead of calling the equivalent of `malloc`.

The first item has to be done entirely in native code because calling `pthread_jit_write_protect_np(false)` from managed code causes a bus error -- this is because `pthread_jit_write_protect_np` controls _every_ page!

## libclrjit.dylib
`libclrjit.dylib` doesn't seem to be one of the modules loaded into the process, which means that the JIT hooks don't work. This is, however, a pretty easy change to make.

## Virtual Method Function Pointers
Hooking virtual methods seems to be a bit fraught (with arm64, at least?) because the function pointers that we get with `DetourRuntimeNETPlatform.GetFunctionPointer` don't seem to related to the entries found in the vtable, and I'm not exactly sure why this isn't the case for amd64...

My quick and dirty solution was to just read the vtable to get the function pointer:
```csharp
var methodTable = method.DeclaringType.TypeHandle.Value;
var methodDesc = method.MethodHandle.Value;

var m_wSlotNumber = (ushort) Marshal.ReadInt16((IntPtr) (methodDesc.ToInt64() + 4));
var m_wFlags = (ushort) Marshal.ReadInt16((IntPtr) (methodDesc.ToInt64() + 6));
if ((m_wFlags & 0x8000) == 0) {
    m_wSlotNumber &= 0xff;
}

var chunkNumber = m_wSlotNumber / 8;
var chunkOffset = m_wSlotNumber % 8;

var vtablePtr = Marshal.ReadIntPtr((IntPtr) (methodTable.ToInt64() + 0x40 + 8 * chunkNumber));
var vtableSlotPtr = (IntPtr) (vtablePtr.ToInt64() + 8 * chunkOffset);

ptr = Marshal.ReadIntPtr(vtableSlotPtr);
```

Additionally, because the devirtualization code expects the entries to point to things that look like precode, I had to change the detour code a bit to match.

### LINKED ISSUE #174: MonoMod Detours fail due to LinuxSystem not supporting ARM
### Description

Hiya, I'm attempting to use pardeike harmony (which in-turn uses MonoMod) to patch a game's headless server. I'm running this under .NET 8 (specifically CLR 8.0.3) on an ARM platform (Oracle's Ampere A1 free tier to be exact) and it appears detours fail to be created. I've tracked it down to the constructor for `MonoMod.Core.Platforms.Systems.LinuxSystem`. Given there's an 'if' statement in there which specifically checks if the platform is x86_64 and then throws a `NotImplementedException`, I get that this is probably just not supported yet. I wanted to report this issue anyways since it's still undesired (if understandable) behavior.

What would need to be done in order for MonoMod to work on ARM under .NET 8?

### Example
This can very easily be reproduced with the following: 

```cs

public static void Main(string[] args)
{
    DetourFactory.Current.CreateDetour(((Delegate)Test).Method, ((Delegate)OtherTest).Method); // Fails with "NotImplementedException" on ARM
    Test();
}


public static void Test()
{
    Console.WriteLine("Test has been called");
}

public static void OtherTest()
{
    Console.WriteLine("Other test has been called");
}

```

LeadAssimilator:
This PR adds Linux and MacOS support for arm64. It is based on a mash up of both [vprodan's](https://github.com/vprodan/MonoMod/tree/reorganize-arm64) and [Fadenfire's](https://github.com/Fadenfire/MonoMod) incomplete existing work along with my own to fully support both platforms from net6.0 on.

Linux support is a straight port of the existing x64 infrastructure to arm64.

MacOS arm64 support is more much more complicated. Unfortunately, dotnet is using Apple's Hardened Runtime which relies on MAP_JIT where pages are W^X. This means to even setup the various thunks/stubs/detours per-thread write protection (!W) must be toggled off/on to perform the write, which simultaneously toggles execution (X) as well. Because jitted code lives on MAP_JIT pages, it can no longer execute during this time which means the write transaction must be performed from preexisting compiled and standard runtime loadable unmanaged code, aka a .dylib. Similarly, none of the dynamic thunks/stubs/detours can execute either. 

The fact that jit code cannot execute when write protection is disabled, has dire implications on the design and implementation of MonoMod.Core which went to great lengths to keep the jit hooks managed. Neither the compileMethod nor allocMem hooks can execute at all with write protection disabled, nor can their wrapping thunks/stubs. Unfortunately, clr jit decides to turn write protection off at various times where a compileMethod invocation can occur. Pre-jitting the hooks generally avoids most of these cases, but not all. Aot might also work, but it isn't available in net6.0 for mac arm64. The only solution is to move the allocMem hook to unmanaged code and split the compileMethod hook into an unmanaged code part and managed code part, where all of the unmanaged code bits live in an unmanaged helper library so they are always executable. Then the unmanaged compileMethod hook part can call the managed part of itself after/post the call to the original compileMethod while controlling or avoiding W^X.


src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,121 @@
...
+    CompileMethod60HookTracker tracker;
+    int entrancy = tracker.entrancy();
+    if (entrancy == 1)
+        memset(&allocMem60_Args, 0, sizeof(allocMem60_Args));
+
+    int res = jitHookConfig60.compileMethod(pThis, comp, info, flags, nativeEntry, nativeSizeOfCode);
+    if (entrancy == 1)
+    {
+        // TODO: Consider instead running this on the same thread but checking the PAL_JitWriteProtect TLS enabledCount var
+        // if (enabledCount > 0) { pthread_jit_write_protect_np(1); .HookPost(...); pthread_jit_write_protect_np(0); }
```
nike4613: Are we sure we can't use some shenanigans like those we do on Linux to check if a known MAP_JIT range is writable for this check?

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
```diff
@@ -23,11 +35,452 @@ public Core60Runtime(ISystem system) : base(system) { }
...
+                                for (var i = 0; i < genericMethodArgs.Length; i++)
+                                {
+                                    genericMethodArgs[i] = JitHookHelpers.GetTypeFromNativeHandle(methodInfo->args.sigInst.methInst[i]).TypeHandle;
+                                }
+                            }
+
+                            var declaringType = JitHookHelpers.GetDeclaringTypeOfMethodHandle(methodInfo->ftn).TypeHandle;
+                            var method = JitHookHelpers.CreateHandleForHandlePointer(methodInfo->ftn);
+
+                            Runtime.OnMethodCompiledCore(declaringType, method, genericClassArgs, genericMethodArgs, (IntPtr)(*nativeEntry), rwEntry, *nativeSizeOfCode);
```
nike4613: Since we're splitting out a `*Post` method anyway (not that I'm super happy about that), we should call the same method from here as well.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,121 @@
...
+extern "C" void* mmch_jit_hook_config(int runtimeMajMin)
+{
+    switch (runtimeMajMin)
+    {
+    case 60:
+        return &jitHookConfig60;
+    default:
+        return nullptr;
+    }
+}
```
nike4613: I think I'd rather have 2 parameters (one major, one minor, possibly even one patch) like we do when creating the `IRuntime` instance on managed side.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,121 @@
...
+    return res;
+}
+
+static void ICorJitInfo_allocMem60_hook(void* pThis, struct AllocMemArgs60* pArgs)
+{
+    jitHookConfig60.allocMem(pThis, pArgs);
+
+    if (compileMethod60_Entrancy == 1)
+        allocMem60_Args = *pArgs;
+}
```
nike4613: Since we're already using C++ for this anyway, I think I'd rather we use templates so that if eventually the signature for `compileMethod` changes, it's not too onerous to add and maintain.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
```diff
@@ -23,11 +35,452 @@ public Core60Runtime(ISystem system) : base(system) { }
...
+
+        [StructLayout(LayoutKind.Sequential)]
+        public struct NativeJitHookConfig
+        {
+            public IntPtr compileMethod;
+            public IntPtr compileMethodHook;
+            public IntPtr compileMethodHookPost;
+            public IntPtr allocMem;
+            public IntPtr allocMemHook;
+        }
```
nike4613: This should probably not be a nested type here, I think. A freestanding type in `MonoMod.Core.Platforms.Runtimes` maybe?

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,121 @@
...
+static int ICoreJitCompiler_compileMethod60_hook(void* pThis, void* comp, void* info, unsigned flags, uint8_t** nativeEntry, uint32_t* nativeSizeOfCode);
+static void ICorJitInfo_allocMem60_hook(void* pThis, struct AllocMemArgs60* pArgs);
+
+static thread_local int compileMethod60_Entrancy = 0;
+static thread_local struct AllocMemArgs60 allocMem60_Args = { 0 };
+static struct JitHookConfig60 jitHookConfig60 =
+{ 
+    .compileMethodHook = &ICoreJitCompiler_compileMethod60_hook,
+    .allocMemHook = &ICorJitInfo_allocMem60_hook
+};
```
nike4613: I don't believe this is standard-conformant C++ until C++20. I don't think it's a huge issue, but it would probably be a good idea to try to keep things properly conformant for the version we're compiling as.

src/MonoMod.Core/Platforms/ISystem.cs
```diff
@@ -61,6 +61,13 @@ public interface ISystem
         /// <param name="data">The data to write into <paramref name="patchTarget"/>.</param>
         /// <param name="backup">A span to fill will the data which was already present, or an empty span.</param>
         void PatchData(PatchTargetKind targetKind, IntPtr patchTarget, ReadOnlySpan<byte> data, Span<byte> backup);
+
+        /// <summary>
+        /// Gets the pointer to the native jit hook configuration struct which can vary by both runtime and arch.
+        /// </summary>
+        /// <param name="runtimeMajMin">Runtime major and minor version.</param>
+        /// <returns>A pointer to the requested jit hook configuration struct.</returns>
+        unsafe IntPtr GetNativeJitHookConfig(int runtimeMajMin);
```
nike4613, line 70: Since this returns a pointer to a type with a well-known structure, we should actually return a pointer to that type, unless Mono doesn't like that, but I don't think Mono has issues with normal pointer types.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
```diff
@@ -23,11 +35,452 @@ public Core60Runtime(ISystem system) : base(system) { }
...
+        private Delegate? ourCompileMethodHookPost;
+
+        private unsafe void InstallNativeJitHook(IntPtr jit)
+        {
+            CheckVersionGuid(jit);
+
+            // Get the real compile method vtable slot
+            var compileMethodSlot = GetVTableEntry(jit, VtableIndexICorJitCompilerCompileMethod);
+
+            var hookConfig = GetNativeJitHookConfig();
```
nike4613: It seems that we never check for `hookConfig` being `null`, even though in several places it returns null to indicate lack of need. This should probably be done above in `InstallJitHook` before deciding to use this codepath.

src/MonoMod.Core/Platforms/ISystem.cs
```diff
@@ -61,6 +61,13 @@ public interface ISystem
         /// <param name="data">The data to write into <paramref name="patchTarget"/>.</param>
         /// <param name="backup">A span to fill will the data which was already present, or an empty span.</param>
         void PatchData(PatchTargetKind targetKind, IntPtr patchTarget, ReadOnlySpan<byte> data, Span<byte> backup);
+
+        /// <summary>
+        /// Gets the pointer to the native jit hook configuration struct which can vary by both runtime and arch.
+        /// </summary>
+        /// <param name="runtimeMajMin">Runtime major and minor version.</param>
+        /// <returns>A pointer to the requested jit hook configuration struct.</returns>
```
nike4613, line 69: This should document that even if the `SystemFeature` flag is set, this may return `null` to indicate that it is not needed.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
```diff
@@ -23,11 +35,452 @@ public Core60Runtime(ISystem system) : base(system) { }
...
+                            {
+                                var allocMemSlot = GetVTableEntry(corJitInfo, Runtime.VtableIndexICorJitInfoAllocMem);
+                                var hookConfig = Runtime.GetNativeJitHookConfig();
+                                hookConfig->allocMem = *allocMemSlot;
+
+                                var ourAllocMemPtr = hookConfig->allocMemHook;
+                                Span<byte> ptrData = stackalloc byte[sizeof(IntPtr)];
+                                MemoryMarshal.Write(ptrData, ref ourAllocMemPtr);
+
+                                Runtime.System.PatchData(PatchTargetKind.ReadOnly, (IntPtr)allocMemSlot, ptrData, default);
```
nike4613, line 390: Why do we do this here, instead of during setup, outside of `CompileMethodHookPost`?

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,121 @@
...
+
+static int ICoreJitCompiler_compileMethod60_hook(void* pThis, void* comp, void* info, unsigned flags, uint8_t** nativeEntry, uint32_t* nativeSizeOfCode)
+{
+    CompileMethod60HookTracker tracker;
+    int entrancy = tracker.entrancy();
+    if (entrancy == 1)
+        memset(&allocMem60_Args, 0, sizeof(allocMem60_Args));
+
+    int res = jitHookConfig60.compileMethod(pThis, comp, info, flags, nativeEntry, nativeSizeOfCode);
```
nike4613: I'm not a fan of this direct call. Exceptions propagating across this may cause issues, particularly with the destructor present. I'd be much more comfortable using the native exception helper even here in native.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,121 @@
...
+
+        hook_post_thread.join();
+    }
+
+    return res;
+}
+
+static void ICorJitInfo_allocMem60_hook(void* pThis, struct AllocMemArgs60* pArgs)
+{
+    jitHookConfig60.allocMem(pThis, pArgs);
```
nike4613: Ditto the exception helper comment here.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
```diff
@@ -23,11 +35,452 @@ public Core60Runtime(ISystem system) : base(system) { }
...
+            allocMemDelegate = CastAllocMemToRealType(CreateAllocMemDelegate());
+            var allocMemFnPtr = EHNativeToManaged(Marshal.GetFunctionPointerForDelegate(allocMemDelegate), out n2mAllocMemHelper);
+
+            // invoke our allocMemFnPtr through IAMP to ensure that the JIT has compiled any needed thunks
+            InvokeAllocMemPtr.InvokeAllocMem(allocMemFnPtr, IntPtr.Zero, null);
+            vtbl[VtableIndexICorJitInfoAllocMem] = allocMemFnPtr;
+        }
+
+        protected virtual int VtableIndexICorJitInfoAllocMem => V60.ICorJitInfoVtable.AllocMemIndex;
+        protected virtual int ICorJitInfoFullVtableCount => V60.ICorJitInfoVtable.TotalVtableCount;
```
nike4613, line 421: Have we verified that AllocMem is present on all builds of .NET 6?

src/MonoMod.Core/Interop/CoreCLR.V60.cs
```diff
@@ -10,12 +10,324 @@ namespace MonoMod.Core.Interop
...
+                >)methodPtr;
+        }
+
         [SuppressMessage("Performance", "CA1812: Avoid uninstantiated internal classes",
             Justification = "It must be non-static to be able to inherit others, as it does. This allows the Core*Runtime types " +
             "to each reference exactly the version they represent, and the compiler automatically resolves the correct one without " +
             "needing duplicates.")]
         public partial class V60 : V50
         {
+            public static class ICorJitInfoVtable
```
nike4613, line 95: Just confirming: was this generated from the source tree of .NET 6, or copied out of .NET 7?

src/MonoMod.Core/Platforms/SystemFeature.cs
```diff
@@ -17,9 +17,15 @@ public enum SystemFeature
...
+
         /// <summary>
         /// This system allows for pages which are protected Read/Execute.
         /// </summary>
         RXPages = 0x02,
+
+        /// <summary>
+        /// This system requires the use of native jit hooks.
+        /// </summary>
+        NativeJitHooks = 0x10,
```
nike4613: Given the actual implementations of the corresponding interface member, this should maybe be called `MayRequireNativeJitHooks`?

src/MonoMod.Core/Platforms/Systems/SystemVABI.cs
```diff
@@ -30,6 +30,32 @@ public static TypeClassification ClassifyAMD64(Type type, bool isReturn)
             return TypeClassification.InRegister;
         }
 
+        public static TypeClassification ClassifyARM64(Type type, bool isReturn)
+        {
+            // ARM64 uses a byref ret buf in x8 if the return size is > 16, but there is no way to model this
+            // currently so just return InRegister for now so the return buffer fixup is effectively disabled
```
nike4613: Out-of-date comment?

src/MonoMod.Core/Utils/AddressMeaning.cs
```diff
@@ -50,9 +56,34 @@ public AddressMeaning(AddressKind kind, int relativeOffset)
...
+        /// <summary>
+        /// Constructs an <see cref="AddressMeaning"/> for the specified <see cref="AddressKind"/>, relative offset and shift amount.
+        /// </summary>
+        /// <param name="kind">The <see cref="AddressKind"/>.</param>
+        /// <param name="relativeOffset">The offset relative to the match start.</param>
+        /// <param name="constantValue">The constant value of the address if kind is <see cref="AddressKind.Constant"/>.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown if <paramref name="kind"/> is invalid
+        /// -OR- <paramref name="kind"/> is absolute
+        /// -OR- <paramref name="relativeOffset"/> is less than zero.</exception>
+        public AddressMeaning(AddressKind kind, int relativeOffset, ulong constantValue)
```
nike4613, line 68: Surely `relativeOffset` is irrelevant if a `constantValue` is provided?

nike4613 (changes requested):
A couple things:
- ApiCompat needs to settle down for tests to run. I'm fine with MonoMod.Core having this kind of breaking change (across minor version bumps, at least, which this will be), so go ahead and `dotnet pack -p:ApiCompatGenerateSuppresionFile=true` and commit the changed ApiCompat suppressions file.
- Given the funkiness around all of the JIT hooks, I'd like an architectural doc in `docs/` that describes both what is being done and why (including the specific constraints that drive it). It's hard enough to make sense of what's going on when you know the codebase as-is.
- I'd like the assembly sources to be a hair more commented than they are. As-is, they're just a pile of assembly that's just there. Even just comments that tell what each section of code is supposed to be doing would be a major improvement.

nike4613:
Oh, and you should probably also enable at least MacOS arm64 test runners. Setting https://github.com/MonoMod/MonoMod/blob/reorganize/build/gen-test-matrix/Models/OS.cs#L105 `Enabled = true` should be enough to do that.

nike4613:
There's also the question of how we can reasonably support Mono JIT hooks, if/when we want to do that. It may be reasonable to just have an entirely separate system for Mono, but something to consider nonetheless.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: You mean with a pipe? Yes, I'm sure - pipes don't work. Whether there exists some other trick, I don't know. I haven't found one yet.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: Yeah, I meant to clean that up, but forgot.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: I thought about doing it that way first, but settled on doing it as a single int so it matches the midfixes/suffixes used everywhere else, like Core21Runtime, Core60Runtime, CoreCLR.V60 and now the native code parts that also use similar conventions. Having maj/min be separate, while being clearer as to which is maj and min, seemed actually more confusing in practice, since you need to resolve everything to the named set of objects using the other midfixes/suffixes convention. I'm basically saying I actually preferred Core60Runtime -> 60 -> jitHookConfig60 over Core60Runtime -> 6, 0 -> jitHookConfig60. Also switching on one number is cleaner ;)

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: Sure, but I think that is a little bit premature. What exactly does or doesn't need shared or templated isn't known until there is actually some change that will require a different implementation. Until that case exists, I think it isn't a good use of time to template it now which would probably have to change later anyway. I can put a TODO there for the next guy though to establish a templated pattern rather than copy pasta.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: The native hook config struct could vary by runtime version which is why it is where it is, following the same pattern as all the other types that can also vary by runtime version.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: Yea it is interesting that clang/gcc allow designated initializers in c++ < 20 and doesn't warn about them unless you -Weverything, probably because they support c99 and allow mixing it with all forms of c++. I'll recompile with -std=c++20.

src/MonoMod.Core/Platforms/ISystem.cs
LeadAssimilator: It is not a well-known structure. It may vary by runtime version.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: I was actually relying on it exploding if its null as it really shouldn't ever be, though perhaps I could have asserted or thrown in various places.

src/MonoMod.Core/Platforms/ISystem.cs
LeadAssimilator: I was actually going for required semantics per the comment as InstallJitHook expects if it is set, there will be a native hook. Though I guess I could see a case where some version of the runtime would need it and others not for a given system and arch.

For optional semantics I don't think we would even need a feature flag at all since InstallJitHook could null check GetNativeJitHookConfig to decide if it should use the native hook or normal managed hooks. But then what should MacOS arm64 GetNativeJitHookConfig do since native jit hooks are required there? Throw? Or should it return null and when null, InstallJitHook additionally checks the required feature flag and throws if set or does nothing?

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: We don't have an ICorJitInfo* outside of an invocation of compileMethod. Unless there is some other way to get one, the vtable patching has to be done during the first invocation. And per the comment above the block, it could be moved into a one-shot Pre method instead of the Post, where the one-shot check and dispatch is done in the unmanaged hook. 

The one-shot is just a perf optimization. If we don't care or think the impact is minimal, it could always get called. Design wise, I prefer symmetry so an always called Pre/Post would be ideal, but given the fact that we would need to use another dedicated thread (or perhaps a thread pool thread) and block on its result like we do Post, I had perf concerns.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: What issues? This is a native->native, c++->c++ call. If an exception is thrown in the real compileMethod, some earlier caller in the clr will catch it. That personality will unwind and the destructor of the tracker will get called. That is primarily why the tracker exists as an object with a destructor, to handle the unwind case without extra glue/hoops. 

If anything, I should consider try/catching the thread start and post invoke.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: Same as above. Unnecessary IMO.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: Yes. AllocMem is actually ancient and is even in netfx. The prototype however changed from separate args to a struct when they added the RW pointers in NET6.

src/MonoMod.Core/Interop/CoreCLR.V60.cs
LeadAssimilator: Both actually. Started life as a cut/paste because all the NET7 logic was moved to NET6 that way and was then edited to conform to the actual NET6 impl.

src/MonoMod.Core/Platforms/SystemFeature.cs
LeadAssimilator: See earlier comments. Considering renaming to RequiresNativeJitHooks so it is clearer, or removing entirely.

src/MonoMod.Core/Platforms/Systems/SystemVABI.cs
LeadAssimilator: No, still valid.

src/MonoMod.Core/Utils/AddressMeaning.cs
LeadAssimilator: I wouldn't say irrelevant, the offset will just get added on, but sure, you could just precompute it as well, so it isn't strictly necessary.

If you don't mind the fact that the ctors will only differ by int vs ulong, but set totally different properties if it is removed, then I will remove it. I generally dislike integral ambiguity like this though in overloads.

src/MonoMod.Core/Platforms/ISystem.cs
nike4613: (I believe) the way I have it with other flags is that the flag indicates that the method *must* be called, but the method can still indicate that no special handling is needed. This way, callers can have an early-out that avoids an interface dispatch by checking the PlatformTriple's cached flags. (This is done for method identification, for instance.)

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
nike4613: The primary issue is that of different C++ runtimes. This usually comes up on Windows, where it is quite common for different DLLs to have linked in different runtimes, but I see no reason why it doesn't apply here; it's not something I want to mess with anyway.

Fundamentally, the EH code is only guaranteed to be able to understand and process its own exceptions--in this case, exceptions that come from C++ code compiled by *this specific* compiler with *these specific* linker args--and in general, we can't match what the runtime uses (and I don't want to try). The exception helper, however, operates on the ABI level and thus *will* mostly-correctly handle any exception passed along the ABI as normal.

src/MonoMod.Core/Interop/CoreCLR.V60.cs
nike4613: For consistency's sake, can you make sure the output here matches what `build/extract-corjit-vtable` outputs?

src/MonoMod.Core/Platforms/Systems/SystemVABI.cs
nike4613: How? It looks like this case should be handled, or at least handleable. What is actually stopping us from implementing this properly?

src/MonoMod.Core/Utils/AddressMeaning.cs
nike4613: To be honest, I'd rather store both in the same `long` field. Then how its treated is determined by the `AddressKind` value, and we can fairly easily validate that.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
nike4613: This winds up running *after* `compileMethod` though, no? We're guaranteed to fail for at least 1 compilation, likely more if there's background JITs happening.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
nike4613: Mmh. Then I'm really not happy with exposing it (even via opaque `IntPtr`) so blatantly on `ISystem`. I'd rather that API return something with *some* well-known structure, even if that isn't a particularly useful one on some platforms.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
nike4613: Poking at the problem with a combination of shitty apple docs and a reasonable LLM, it seems like the only option is signal handling, but of course that is a *massive* problem since signals are process-global. It did bring up some functionality that I was not aware of though: signal blocking. *In principal*, it might be possible to safely:

1. `sigprocmask()` (process-wide) block `SIGBUS` and `SIGSEGV`
2. install our own handlers to pick up those
3. `pthread_sigmask()` to (thread-local) unblock `SIGBUS` and `SIGSEGV` (at this point, ONLY this thread should be able to raise those two, so the global handlers are fine)
4. `sigsetjmp` (so we can resume execution in case of exception) then test-write into a known `MAP_JIT` region
5. then cleanup backwards: restore local-thread blocks (so re-block sigbus and segv), install old handlers, restore process-wide blocks (unblocking sigbus and segv again)

Of course, this would have to be locked around to ensure that it doesn't race against itself (and then hope that it's not racing against someone else trying to do the same thing), but it might just work.

src/MonoMod.Core/Platforms/ISystem.cs
LeadAssimilator: Ok, if that is the pattern then I guess the flag can stay and I'll update the comment. But what about the other questions? How should we handle the case where the native jit hook is actually required and we don't want to fall back to the managed hooks? Should we use another flag for that, or should GetNativeJitHookConfig throw? and if it throws, should it be caught/logged and ignored leaving the or is fatal? We can't do anything if it fails, but it is unclear if we should crash the process or continue with no hooks working.

nike4613: I'd prefer throwing, as we cannot provide our normal guarantees. Better an explosive failure than a silent one here.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: I think you should make peace with the almighty IntPtr :)
There really cannot be a well known struct that covers all future runtime hook needs. Even if you choose to use a single struct and modify it, you have broken the contract. And ISystem doesn't ever care what the native jit hook config struct layout is, it just needs to pass it through to CoreXXRuntime from the system specific native helper, so something opaque makes the most sense. Since only CoreXXRuntime and the unmanaged native helper must share the same contract, the responsibility for converting the opaque pointer to something concrete lives within each, which is similar to the casting of the delegates in CoreXXRuntime or even INativeExceptionHelper that is also playing fast and loose with IntPtrs and types. If you are mucking with a native jit hook, you already need to know all this, and the extra type safety you are wishing for wont actually provide any meaningful value. Honestly, who is going to implement another ISystem anyway and be bothered by this detail? Complete n00bs can't play here, so I don't think this is anything to worry about, given the target audience.

If you are concerned about the method's public visibility in ISystem, then perhaps there should be a separate internal interface with it so it remains hidden and CoreXXRuntime could is test for it.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: Fail? no. Will miss them? yes. Yes, it happens after, and I stated in the TODO comment that it may be necessary to force a compileMethod invocation to prime things or the pre hook way needs done. In practice, something always needs compiled so I think the probability we miss an important one is low. Background jit could already occur prior to the compileMethod hook install anyway, so I'm not that concerned about it specifically.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: Remember, the helper is platform specific. dotnet on macos links against the system libc++ just like the helper, so there is no difference.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
LeadAssimilator: Signals suck. Messing with them can lead to a world of hurt. Mach has native support for hardware exceptions via an exception port which is probably a better option. I assume it has priority anyway. I don't really like either option though since they add complexity, can introduce hard to diagnose bugs in the host app and are slow.

Dumping the impl of pthread_jit_write_protect_np shows it is changing the contents of a system register. I've modified the helper to read it and act accordingly. IMO this is way simpler and less error prone. The only downside is it may change with the OS or cpu, though it hasn't yet for years and I would expect if it were to change, that it would be in a new major release anyway since it has both entrenched user and kernel components.

src/MonoMod.Core/Interop/CoreCLR.V60.cs
LeadAssimilator: I would if only it wasn't broken. I think clang sharp is returning bunk results, namely CXXMethodDecl.IsVirtual is false when it should clearly be true. I manually double checked the indexes/counts though and they are accurate, namely because I can read and count better than clang sharp can apparently.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
nike4613: We'll see how stable that ends up being, I suppose. It did also occur to me that since the runtime tracks the *number* of calls asking to set and unset write protection, we need not only to restore the write protection on exit of the jit hook, but also as soon as the current thread re-enters the JIT, otherwise it won't correctly unset protection if it needs to. (And this will probably eventually cause issues if/when the JIT starts calling managed methods, for the same reason, but we'll revisit that when we get there.)

I'm increasingly convincing myself that the correct approach here is PDB introspection...

src/MonoMod.Core/Interop/CoreCLR.V60.cs
LeadAssimilator: Oh and just fyi both v and s dtors are at the end for *nix and just v on win - nothing is taking that into consideration afaik, so technically the overall buffer is allocated 1 or 2 pointers worth too small. Granted things would be real bad if the wrapped vtable were destroyed even if those slots were there, so this isn't a problem in practice.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
nike4613: That could be a failure, as the *very next* compilation *could* be one we care about; I believe I can even write an application where that is the case always. Ideally we don't miss *any*. There shouldn't be too much problem inserting a managed `*Pre` method to set this up, should there?

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
nike4613: I feel like that is still prone to issues. I admit I don't know Apple's support policy very well, but what if, say:

- Apple releases MacOS 17, which includes a breaking libc++ change. This breaks all existing runtimes, including our helper.
- .NET 12 releases supporting MacOS 17, compiling against the new libc++.

What do we do?

We still need to support .NET 6, on the old OS and libc++, while also supporting the new ones. As best I understand, doing this will *hard require* multiple helper binaries to do, which is very strongly undesirable. I think we're safer from that with the exception helpers.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
```diff
@@ -106,6 +107,12 @@ private unsafe void InstallNativeJitHook(IntPtr jit)
             MemoryMarshal.Write(ptrData, ref ourCompileMethodHookPtr);
 
             System.PatchData(PatchTargetKind.ReadOnly, (IntPtr)compileMethodSlot, ptrData, default);
+
+            // now trigger the compilation of something to ensure that any patches deferred to first invoke are performed
+            var dm = new DynamicMethod("CompileMethodPatchPrimer", typeof(void), null, typeof(Core60Runtime), false);
+            var il = dm.GetILGenerator();
+            il.Emit(OpCodes.Ret);
+            dm.Invoke(null, null);
```
nike4613: I think we can (and probably should) use a single-callsite `NoInlining` method for this, or at least use a DMD here instead of DynamicMethod directly.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: I've fixed this for now by immediately compiling a dynamic method and invoking it right after installing the compileMethod hook. Adding a Pre is not good if we have to fall back to the separate thread dispatch unless it is a one-shot Pre aka PreInit or something.

Unless there is some major issue with this, I think it makes sense to just release the current impl in preview and then revisit after user testing/feedback. If the system register read method proves to be stable or it gets replaced with some sym lookup to find that enabledCount tls var PAL is using, then an always called Pre should be fine.

nike4613 (changes requested):

nike4613:
All the ARM test runs failed :(

LeadAssimilator:
> All the ARM test runs failed :(

The hardened runtime is apparently turned off or unavailable in the github action containers and trying to read the system register results in a crash. It is working now after adding the relevant gaurds in the helper. All tests passed with it that way. I've since added a global check to bypass the native jit hooks entirely and use the managed hooks in that case, though perhaps you may actually want to always use the native hooks, so there is at least some automated test coverage of that path, but the important write protect bits unfortunately can't be tested.

src/MonoMod.Core/Platforms/Runtimes/Core60Runtime.cs
LeadAssimilator: done.

nike4613:
I remember seeing somewhere (I don't remember where exactly) that the `dotnet` apphost doesn't have hardened runtime enabled, but that the normal apphosts do. I'd like to test with hardened runtime on, as well as with it off, though I'm not sure how we can actually do that. Any ideas there?

Also seems like somehow we broke Mono on all MacOS?

src/MonoMod.Core/MonoMod.Core.csproj
```diff
@@ -23,6 +23,9 @@
   <ItemGroup>
     <EmbeddedResource Include="Platforms\Architectures\x86_64\exhelper_linux_x86_64.so" LogicalName="exhelper_linux_x86_64.so" />
     <EmbeddedResource Include="Platforms\Architectures\x86_64\exhelper_macos_x86_64.dylib" LogicalName="exhelper_macos_x86_64.dylib" />
+    <EmbeddedResource Include="Platforms\Architectures\arm64\exhelper_linux_arm64.so" LogicalName="exhelper_linux_arm64.so" />
+    <EmbeddedResource Include="Platforms\Architectures\arm64\exhelper_macos_arm64.dylib" LogicalName="exhelper_macos_arm64.dylib" />
+    <EmbeddedResource Include="Platforms\Architectures\arm64\helper_macos_arm64.dylib" LogicalName="helper_macos_arm64.dylib" />
```
nike4613: I'd still like to link these together into one binary.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,210 @@
...
+                hook_post_thread.join();
+#else
+                PthreadJitWriteProtectScope jwps(true);
+                res = jitHookConfig.compileMethodHookPost(pThis, comp, info, flags, nativeEntry, nativeSizeOfCode, res, &allocMem_Args);
+#endif
+            }
+            catch (...)
+            {
+                // Just catch and eat all exceptions as we don't want anything making their way back to the compileMethod callers
+            }
```
nike4613: I don't think there are any exceptions that can happen that this can catch. Managed exceptions will assert when trying to cross the native-managed boundary, and protection violations can't be caught.

src/MonoMod.Core/Platforms/Architectures/arm64/helper_macos_arm64.cc
```diff
@@ -0,0 +1,210 @@
...
+
+extern "C" void mmch_jit_memcpy(void* dst, const void* src, size_t n)
+{
+#if defined(DISABLE_JWP_DETECTION)
+    int jwp = pthread_jit_write_protect_supported_np();
+    if (jwp)
+        pthread_jit_write_protect_np(0);
+#else
+    PthreadJitWriteProtectScope jwps(false);
+#endif
```
nike4613: Surely we should always be using `PthreadJitWriteProtectScope` here? No need to `#define` it out.

nike4613 (changes requested):

Lord-Kamina:
Just a question for all you folks working on this...

From what I understand, this is targeting .NET Core 6+ and that would mean it won't work with apps that are still using .NET framework, yes?

Is there any particular technical reason for this or is it just because .NET framework is really old?

nike4613:
I believe this should work as-is for .NET Framework on Windows ARM64; the only thing it might be missing is the precode. Most of the effort here is working around MacOS-specific quirts.

I think GitHub Actions now actually provides Windows ARM64 runners-- @LeadAssimilator Can you add those to CI and see if .NET Framework does actually work? If not, we can fix that in a followup I think.

Lord-Kamina:
I'm asking for example in the context of people using HarmonyX/BepInEx which relies on MonoMod for modding Unity games. I think Unity is using either .NET framework 4.x or .NET Standard 2.x. My actual question is whether this will work with these runtimes when it eventually lands?

ETA: I'm talking about people playing Unity games (for example Valheim) on Apple Silicon.

nike4613:
This should work there, yes. Look for the Mono test runs; that's what Unity uses (though we don't actually test the Unity builds of Mono at the moment).

src/MonoMod.Core/Platforms/Systems/MacOSArm64Helper.cs
```diff
@@ -0,0 +1,140 @@
...
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(tmplName);
+            }
+        }
+
+        private static void ExtractEmbeddedResourcesToTempFile(int fd, string logicalName)
+        {
+            using (var handle = new SafeFileHandle((IntPtr)fd, true))
+            using (var stream = new FileStream(handle, FileAccess.Write))
```
I-Info: `SafeFileHandle` will not work properly with P/Invoked file descriptor on Mono runtime. 
See: mono/mono/issues/12783

pardeike:
@LeadAssimilator  Would you mind adding the changes from [this commit](https://github.com/pardeike/MonoMod/commit/e7d8663b7b8447ede7f330bade75e6ea95866683?w=1) to this PR?

src/MonoMod.Core/Platforms/Systems/SystemVABI.cs
nike4613: The bug was in the ABI fixup, which didn't correctly handle the case where the return buffer is not passed as a normal argument. This has been fixed, and the implementation is whole.

.github/workflows/run-with-dumps.ps1
```diff
@@ -42,14 +42,14 @@ elseif ($IsLinux -or $IsMacOS)
...
 # on MacOS, SIGXCPU doesn't coredump by default. Thus, we use LLDB unattended to perform the dump 
 # because we run our Linux stuff in containers, we can't set the core_pattern. Thus, we'll do the same thing we *must* do on MacOS and use LLDB to generate dumps when crashing
-xargs lldb -x -b -O "command alias sdmp process save-core -p minidump -s full -- '$(Join-Path $dumpsPath 'dump_crash.core')'"\
+xargs lldb -x -b \
+    -O "command alias sdmp process save-core -s full -pminidump '$(Join-Path $dumpsPath 'dump_crash.core')'"\
     -s "$(Join-Path $lldbHelpers 'setup.lldb')" \
     -K "$(Join-Path $lldbHelpers 'crash.lldb')" \
-    -s "$(Join-Path $lldbHelpers 'teardown.lldb')" -- "$Exe";
+    -s "$(Join-Path $lldbHelpers 'teardown.lldb')" -- timeout -s XCPU -k 60 300 "$Exe";
 exit `$?;
```
nike4613, line 53: This whole mess needs to be replaced with python scripting unfortunately, but that's a whole other task.

nike4613 (approved):

PR #258, opened 2025-09-10
Support .NET 10

TianMengLucky:


nike4613:
The Windows failures all appear to be https://github.com/dotnet/runtime/issues/119566, but the others appear legitimate. Seems like they finally changed the structure of the runtime assembly structures.

build/gen-test-matrix/Models/Dotnet.cs
```diff
@@ -111,6 +111,17 @@ internal sealed record Dotnet : Enableable
             MonoPackageSource = Constants.NuGetSource.NugetOrg,
             MonoPackageVersion = "9.0.0",
 
+            RIDs = ["win-x86", "win-x64", "win-arm64", "linux-x64", "linux-arm", "linux-arm64", "linux-musl-x64", "linux-musl-arm", "linux-musl-arm64", "osx-x64", "osx-arm64"]
+        },
+        new()
+        {
+            Name = ".NET 10.0 RC",
```
nike4613: I think I'd prefer to name this just `.NET 10.0`, and we'll just keep it updated as we support the later RCs and eventually GA.

nike4613 (changes requested):

nike4613:
What is suddenly including extra nullable attributes?

TianMengLucky:
Strange, there is a conflict between `MonoMod/external/iced/src/csharp/Intel/Iced/NRT.cs` and the `Nullable` reference in `Common.CS.props`.

nike4613:
You didn't update iced, did you? I don't see it in the diff. What changed to trigger that?

TianMengLucky:
I tested locally, but don't know what caused it.

nike4613:
Is the ReproducibleBuilds update maybe the problem?

TianMengLucky:
> Is the ReproducibleBuilds update maybe the problem?

No

TianMengLucky:
The issues currently encountered only appear after upgrading the SDK version to RC2.

nike4613 (approved):

PR #263, opened 2025-10-06
Fix PlatformDetection calling DynDll before it's initialized on linux

js6pak:


js6pak:
@nike4613 Do you have any idea why this test failure happens only on macos-13? Should we even be trying `libdl.so.2` on macos?

nike4613:
I do not, but I also haven't looked into it. I don't think it should be tried here either, nor should we be hitting that codepath.

nike4613:
Looks like it's Mono on MacOS specifically that's failing here

js6pak:
And even more specifically macos-x64, and I can't reproduce it under rosetta

nike4613:
CI is showing it on 14/15 both ARM and Intel

nike4613 (approved):

PR #264, opened 2025-10-06
Make DetourFactory.Current mutable

js6pak:


src/MonoMod.Core/IDetourFactory.cs
```diff
@@ -83,26 +83,38 @@ public readonly record struct CreateNativeDetourRequest(IntPtr Source, IntPtr Ta
...
     {
+        private static IDetourFactory? current;
+
+        /// <summary>
+        /// Gets or sets the current <see cref="IDetourFactory"/>.
+        /// </summary>
+        public static IDetourFactory Current
+        {
+            get => current ?? Default;
+            set => current = value;
```
nike4613: I'm not a huge fan of this API shape. I think my preference would be something that looks instead like:

```cs
public static IDetourFactory Current { get; }

[EditorBrowsable(Never)]
public static void SetCurrentFactory(Func<Action<IDetourFactory>, IDetourFactory> creator);
```

Used like:

```cs
DetourFactory.SetCurrentFactory(getCur => new MyFactory(getCur()));
```

nike4613 (changes requested):

src/MonoMod.Core/IDetourFactory.cs
js6pak: I thought about this when writing it and I don't think it matters that much. I'm fine with either.
Is EB.Never really necessary though?

nike4613: I think EB(Never) is a good idea, since this is something that *really* should not be set by the vast, vast majority of users of the library. I also think it should be more explicit than a property setter, hence the method (and the delegate mess lets us not initialize the default factory if it's not needed for the new one, though that may not really be necessary).

nike4613 (approved):

PR #265, opened 2025-10-06
Fix native hooks on macos

js6pak:


js6pak:
Note the macos-x64 failure is not new, there just wasn't a native hook test before.

PR #280, opened 2025-12-10
Publicize RuntimeDetour extensions

Meivyn:
Every extension class is public but this one, and I do have a use for this method.


nike4613 (commented):
Part of me thinks it would be better for this to be moved into Utils

PR #284, opened 2026-01-30
Add DllPrefix/DllSuffix utilities

js6pak:


nike4613 (approved):
I'm OK with leaving these without XML docs for now, as they seem reasonably self-explanatory. I don't imagine they would be too much to add here though.

nike4613 (approved):

PR #285, opened 2026-01-30
Add ISystem.EnumerateLoadedModules

js6pak:


src/MonoMod.Core/Interop/StringParser.cs
```diff
@@ -0,0 +1,331 @@
+// https://github.com/dotnet/runtime/blob/48a162cff4bdfc77d7d7767497b073481febf1d8/src/libraries/Common/src/System/IO/StringParser.cs
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using MonoMod.Utils;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace System.IO
```
nike4613: We should probably change this namespace, no? Esp. given the attribution comments above.

src/MonoMod.Core/Platforms/Systems/LinuxSystem.cs
```diff
@@ -21,9 +19,20 @@ internal sealed class LinuxSystem : ISystem, IInitialize<IArchitecture>
...
+            }
+        }
+
         public IEnumerable<string?> EnumerateLoadedModuleFiles()
         {
-            return Process.GetCurrentProcess().Modules.Cast<ProcessModule>().Select(m => m.FileName)!;
+            foreach (var module in EnumerateLoadedModules())
+            {
+                yield return module.FileName;
+            }
```
nike4613, line 35: Would it not be better to avoid the extra allocs of LoadedModule instances here?

src/MonoMod.Core/Interop/StringParser.cs
js6pak: I thought that considering I'm basically vendoring the file as is, I might as well keep the namespace. I could move it to `MonoMod.Core.Utils` though?

src/MonoMod.Core/Platforms/Systems/LinuxSystem.cs
js6pak: I wanted to keep `EnumerateLoadedModuleFiles` impl the same on all systems (and ideally make it a DIM but oh well). Making `LoadedModule` a struct might be nicer here?

nike4613: Is there a reason its not a struct in the first place?

src/MonoMod.Core/Interop/StringParser.cs
nike4613: Surely vendoring the file includes the namespace

nike4613 (approved):

PR #288, opened 2026-01-31
Add libgcc_s.so.1 to exhelper_linux_x86_64.so's DT_NEEDED

js6pak:
This should help with resolving `_Unwind_RaiseException` in custom hosts

TODO figure out whether this will break when the runtime is compiled with compiler-rt and/or if there is a better way to solve this


nike4613:
I'm actually kinda shocked this doesn't die horribly on musl. I suppose its libgcc, not glibc. This does still seem like something that's fairly fragile though.

PR #293, opened 2026-02-27
Allow methodbuilder dmd generator for net9+

isolno:
With Net9 introducing PersistedAssemblyBuilder as a way to save AssembyBuilders, there is good reason to support it by allowing the 'methodbuilder' DMDType. For my use case, the 'cecil' generator is over 10 times slower, and with many failures, making it not actually usable. While methodbuilder is more stable and interactable.

### DIFF
--- a/tools/Common.CS.targets
+++ b/tools/Common.CS.targets
@@ -104,6 +104,11 @@
         <CecilVersion Condition="'$(CecilVersion)' == ''">$(NewCecilVersion)</CecilVersion>
         <NETShimVersion>4.*</NETShimVersion>
       </PropertyGroup>
+      
+      <!-- MethodBuilder-->
+      <PropertyGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', '9.0'))">
+        <DefineConstants>METHODBUILDER_SUPPORTED;$(DefineConstants)</DefineConstants>
+      </PropertyGroup>
 
       <ItemGroup>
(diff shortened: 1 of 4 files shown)


tools/Common.CS.targets
```diff
@@ -104,6 +104,11 @@
         <CecilVersion Condition="'$(CecilVersion)' == ''">$(NewCecilVersion)</CecilVersion>
         <NETShimVersion>4.*</NETShimVersion>
       </PropertyGroup>
+      
+      <!-- MethodBuilder-->
+      <PropertyGroup Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', '9.0'))">
+        <DefineConstants>METHODBUILDER_SUPPORTED;$(DefineConstants)</DefineConstants>
+      </PropertyGroup>
```
nike4613: I think I'd rather `#define` this in the 2 files that need it with the appropriate already-defined conditionals.

nike4613 (changes requested):
I certainly don't have a problem with doing this, though I might ask: what has you falling out to Cecil in the first place? Generally, we try to use the DynamicMethod backend for everything.

isolno:
Im using it to save dmds and ilhooks to an assembly, and DMDEmitMethodBuilderGenerator.GenerateMethodBuilder makes this easy. Not meant to replace dynamicmethod for common use.

nike4613 (approved):

PR #298, opened 2026-03-10
Fix JIT compilation for methods on generic types

JBou:
Until now, I had to stay on `Harmony 2.2.2`, because patches on generic types where not working with `Harmony 2.3` using `MonoMod.Core` anymore. This PR would allow to patch them again. I verified in our production app, and am able to apply the patches again.

### Changes:
Update `FxCoreBaseRuntime.Compile` to pass generic type arguments to `RuntimeHelpers.PrepareMethod` when the method is declared on a generic type. This ensures that the runtime prepares the correct method instantiation.

Add unit tests `Test_GenericClassDetour` and `Test_GenericClassDetourWithoutGenericParameter` to `DetourTest` to verify detours on generic types.

### DIFF
--- a/src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
+++ b/src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
@@ -374,7 +374,16 @@ namespace MonoMod.Core.Platforms.Runtimes
         public virtual void Compile(MethodBase method)
         {
             var handle = GetMethodHandle(method);
-            RuntimeHelpers.PrepareMethod(handle);
+            if (method.DeclaringType?.IsGenericType == true)
+            {
+                var type = method.DeclaringType;
+                var genericParams = type.GetGenericArguments().Select(x => x.TypeHandle).ToArray();
+                RuntimeHelpers.PrepareMethod(handle, genericParams);
+            }
+            else
+            {
+                RuntimeHelpers.PrepareMethod(handle);
+            }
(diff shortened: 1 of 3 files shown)


src/MonoMod.UnitTest/RuntimeDetour/DetourTest.cs
```diff
@@ -90,6 +91,69 @@ public void TestDetours()
...
+
+            var replacement1 = typeof(TestObjectPatch).GetMethod("TestDetour1");
+            Assert.NotNull(replacement1);
+
+            var obj1 = new TestObjectDouble();
+            var obj2 = new TestObjectString();
+            Assert.Equal(1, obj1.GetDouble());
+            Assert.Equal(1, obj2.GetDouble());
+
+            var detour1 = DetourFactory.Current.CreateDetour(originalMethod1, replacement1);
```
nike4613, line 108: This should be in a `using` block.

src/MonoMod.Core/Platforms/Runtimes/FxCoreBaseRuntime.cs
nike4613, line 386: If we're going to try to do any handling of generics, it should be complete, and deal with method instantiations properly as well. As such, the generic type param list construction should probably be its own helper (maybe in utils? There may also already be an appropriate helper somewhere in the JIT hook logic.)

nike4613 (changes requested):
We do not support patching generics at all at the moment, and while I'm perfectly willing to take improvements in this area, this is an extremely difficult area, and I doubt this fixes the fundamental problem.

PR #301, opened 2026-03-17
fix abi fixup

wuke32767:
when i'm implementing generic detour, i encountered several abi fixup issues.

generic detour is not finished, but i want to merge this thing first.

test result for "old" monomod: https://github.com/wuke32767/MonoMod/actions/runs/23201287457

### DIFF
--- a/src/MonoMod.Core/Platforms/PlatformTriple.cs
+++ b/src/MonoMod.Core/Platforms/PlatformTriple.cs
@@ -673,7 +673,7 @@ namespace MonoMod.Core.Platforms
             var returnType = fromInfo.ReturnType;
             var hasReturnBuffer = Abi.Classify(returnType, true) is TypeClassification.ByReference;
             var hasThis = !fromInfo.IsStatic;
-            var requiresReturnBufferFixup = hasThis && toInfo.IsStatic && hasReturnBuffer && returnBufferIsArgument;
+            var requiresReturnBufferFixup = hasThis == toInfo.IsStatic && hasReturnBuffer && returnBufferIsArgument;
 
             // Whenever we detour a call from a generic method, depending on the ABI, we may
             // receive a generic context as an argument, which a callee never needs, at least
@@ -709,9 +709,11 @@ namespace MonoMod.Core.Platforms
                 // Thus, no ABI fixups are needed, return `to` as is.
                 return to;
             }
+            // switch to a dynamic method. toInfo must be refreshed.
+            requiresReturnBufferFixup = hasThis == /*toInfo.IsStatic*/true && hasReturnBuffer && returnBufferIsArgument;
 
-            var returnBufferType = hasReturnBuffer ? returnType.MakeByRefType() : returnType;
-            var newReturnType = hasReturnBuffer && !Abi.ReturnsReturnBuffer ? typeof(void) : returnBufferType;
+            var returnBufferType = requiresReturnBufferFixup ? returnType.MakeByRefType() : returnType;
+            var newReturnType = requiresReturnBufferFixup && !Abi.ReturnsReturnBuffer ? typeof(void) : returnBufferType;
 
             var thisPos = -1;
             var returnBufferPos = -1;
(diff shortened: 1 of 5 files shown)


src/MonoMod.Core/Platforms/PlatformTriple.cs
```diff
@@ -673,7 +673,7 @@ public MethodBase GetRealDetourTarget(MethodBase from, MethodBase to)
             var returnType = fromInfo.ReturnType;
             var hasReturnBuffer = Abi.Classify(returnType, true) is TypeClassification.ByReference;
             var hasThis = !fromInfo.IsStatic;
-            var requiresReturnBufferFixup = hasThis && toInfo.IsStatic && hasReturnBuffer && returnBufferIsArgument;
+            var requiresReturnBufferFixup = hasThis == toInfo.IsStatic && hasReturnBuffer && returnBufferIsArgument;
```
nike4613, line 676: I don't understand the logic behind this change. If `from` is static, and `to` is instance, we need this fixup? does the fixup even correctly account for static->instance detours?

src/MonoMod.Core/Platforms/PlatformTriple.cs
```diff
@@ -709,9 +709,11 @@ public MethodBase GetRealDetourTarget(MethodBase from, MethodBase to)
                 // Thus, no ABI fixups are needed, return `to` as is.
                 return to;
             }
+            // switch to a dynamic method. toInfo must be refreshed.
+            requiresReturnBufferFixup = hasThis == /*toInfo.IsStatic*/true && hasReturnBuffer && returnBufferIsArgument;
```
nike4613: Why do we also change the logic here? What about the above checks changes what's correct?

src/MonoMod.Core/Utils/BytePatternCollection.cs
```diff
@@ -479,8 +481,21 @@ public bool TryFindMatch(ReadOnlySpan<byte> data, Span<byte> addrBuf, [MaybeNull
...
+            get
+            {
+                if (!lazyPossibleFirstBytesInitialized)
+                {
+                    lazyPossibleFirstBytes = GetPossibleFirstBytes();
+                    lazyPossibleFirstBytesInitialized = true;
+                }
+                return lazyPossibleFirstBytes;
+            }
+        }
```
nike4613, line 498: I don't think this is actually appreciably more threadsafe than the old code.

src/MonoMod.Core/Platforms/PlatformTriple.cs
wuke32767: the reason why static->instance detours requires abi fixup, is the same as why instance->static detours requires abi fixup.
you may think we got static->dmd(static)->instance in the end, where dmd seems to be doing nothing. imo as long as A can't be directly detoured to B, fixup is needed.

src/MonoMod.Core/Platforms/PlatformTriple.cs
wuke32767: i seems to understand what did the original author is thinking, so i replemented it.
but... as a result, https://github.com/MonoMod/MonoMod/pull/301/changes/00f60b4795509b37edcb03c3c3935e603cd5e99b

src/MonoMod.Core/Utils/BytePatternCollection.cs
wuke32767: you may be thinking two threads setting value at the same time is not safe. however, calling GetPossibleFirstBytes for multiple times gives the exactly same result (except for one array reference), so it's safe enough.
for any other issue, https://learn.microsoft.com/en-us/archive/msdn-magazine/2012/december/csharp-the-csharp-memory-model-in-theory-and-practice#c-memory-model-according-to-ecma-334

nike4613: How is this actually different from the original code though? a `Nullable<T>` is functionally a `T` and a `bool` anyway. Plus `volatile` doesn't introduce any memory barriers, it only guarantees that the write won't be cached or coalesced by the runtime.

nike4613: And given that this pattern exists elsewhere without issues, unless the original manifests problems in some case (in which case those other places should be fixed too) this should be reverted IMO.

wuke32767: volatile fields do provide acquire-­release semantics in c#
i also want to know why it wasn't an issue until today

nike4613: Do you have a specific example of where this causes issues?

nike4613: In any case, we have threaded initialization helpers which should be used instead of this.

PR #302, opened 2026-03-20
Add Test for and fix #299

### LINKED ISSUE #299: DMDEmitMethodBuilderGenerator crashes Mono when saving generated assembly
### Description

Saving methods generated by `DMDEmitMethodBuilderGenerator` on Mono crashes Mono due to passing empty array to `DefineMethod` type modifiers parameters instead of nulls

Mono crashes with `* Assertion at sre-encode.c:290, condition 'count > 0' not met`

MonoMod logs:
```
Detected runtime: Mono 6.8.0.105 using Framework corelib
```

### Examples
```cs
var assemblyName = new AssemblyName("Example");
var assemblyBuilder = AppDomain.CurrentDomain.DefineDynamicAssembly(assemblyName, AssemblyBuilderAccess.RunAndSave);
var dynamicModule = assemblyBuilder.DefineDynamicModule(assemblyName.Name, assemblyName.Name + ".dll");
var typeBuilder = dynamicModule.DefineType("Test");

var dmd = new DynamicMethodDefinition("Test", typeof(void), []);

var il = dmd.GetILGenerator();
il.Emit(OpCodes.Ret);

DMDEmitMethodBuilderGenerator.GenerateMethodBuilder(dmd, typeBuilder);

typeBuilder.CreateType();
assemblyBuilder.Save("Example.dll");
```

```cs
Switches.SetSwitchValue("DMDType", "methodbuilder");
Switches.SetSwitchValue("DMDDumpTo", "Example");
DynamicMethodDefinition dmd = new("Test", typeof(void), []);

var il = dmd.GetILGenerator();
il.Emit(OpCodes.Ret);

dmd.Generate();
```

Example of working DefineMethod call
```cs
typeBuilder.DefineMethod(
    "Test", 
    MethodAttributes.Public | MethodAttributes.Static, 
    CallingConventions.Standard,
    typeof(void), null, null, [], [], []
);
```
and example of crashing DefineMethod call
```cs
typeBuilder.DefineMethod(
    "Test", 
    MethodAttributes.Public | MethodAttributes.Static, 
    CallingConventions.Standard,
    typeof(void), [], [], [], [], []
);
```
do note that parameter modifiers are also affected, so this works
```cs
typeBuilder.DefineMethod(
    "Test", 
    MethodAttributes.Public | MethodAttributes.Static, 
    CallingConventions.Standard,
    typeof(void), 
    null, null, 
    [typeof(int)], /* parameter types */
    [null], [null]
);
```
and this crashes
```cs
typeBuilder.DefineMethod(
    "Test", 
    MethodAttributes.Public | MethodAttributes.Static, 
    CallingConventions.Standard,
    typeof(void), 
    null, null, 
    [typeof(int)], /* parameter types */
    [[]], [[]]
);
```

### Fix suggestion

I suggest making `_DMDEmit.ResolveWithModifiers` return nulls instead of empty arrays on Mono

xADDBx:
Screenshot of the test actually failing on a mono runtime:
grafik
Screenshot of it passing on the same runtime after the fix:
grafik

Relevant mono snippet: https://github.com/mono/mono/blob/0f53e9e151d92944cacab3e24ac359410c606df6/mono/metadata/sre-encode.c#L290

Fixes #299

### DIFF
--- a/src/MonoMod.UnitTest/DynamicMethodDefinitionTest.cs
+++ b/src/MonoMod.UnitTest/DynamicMethodDefinitionTest.cs
@@ -1,4 +1,4 @@
-﻿extern alias New;
+extern alias New;
 
 using Mono.Cecil.Cil;
 using MonoMod.Cil;
@@ -324,11 +324,11 @@ namespace MonoMod.UnitTest
                 var methodBuilder = DMDEmitMethodBuilderGenerator.GenerateMethodBuilder(dmd, typeBuilder);
                 Assert.NotNull(methodBuilder);
                 var methodBuilder2 = DMDEmitMethodBuilderGenerator.GenerateMethodBuilder(dmd2, typeBuilder);
-                Assert.NotNull(methodBuilder);
+                Assert.NotNull(methodBuilder2);
                 var methodBuilder3 = DMDEmitMethodBuilderGenerator.GenerateMethodBuilder(dmd3, typeBuilder);
-                Assert.NotNull(methodBuilder);
+                Assert.NotNull(methodBuilder3);
                 var methodBuilder4 = DMDEmitMethodBuilderGenerator.GenerateMethodBuilder(dmd3, typeBuilder);
-                Assert.NotNull(methodBuilder);
+                Assert.NotNull(methodBuilder4);
 
                 var generatedType = typeBuilder.CreateType();
                 Assert.NotNull(generatedType);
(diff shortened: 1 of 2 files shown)


xADDBx:
Alternatively this could be fixed by putting the check directly in ResolveWithModifiers as was mentioned in the original issue.
grafik

(Though that probably should also include a change in method signature (Type[]?)

PR #304, opened 2026-04-08
Fix #282

### LINKED ISSUE #282: DMD Generate overrides modified parameters with originals
### Description

When generating from DynamicMethodDefinition, if it has a non-null OriginalMethod, it will ignore the parameters of the new method definition, instead using the originals. While there is a work around of setting OriginalMethod to null before generating, that can remove useful information, and requires using reflection to set a compiler generated field. Issue happens when generating with 'dynamicmethod' or 'methodbuilder', but not 'cecil'.

### Example
```cs
[Fact]
public void DMDGenerateDoesNotOverrideParameters()
{
  using var dmd = new DynamicMethodDefinition(((Delegate)func).Method);
  // modify to func(a, b) => a + b;
  dmd.Definition.Parameters.Add(new(dmd.Module.TypeSystem.Int32));
  using (var il = new ILContext(dmd.Definition))
  {
    var c = new ILCursor(il);
    c.Index += 1;
    c.Emit(OpCodes.Ldarg_1);
    c.Emit(OpCodes.Add);
  }
  
  Assert.Equal(1, func(1));
  Test("cecil");
  Test("dynamicmethod");
#if NETFRAMEWORK
  Test("methodbuilder");
#endif

  static int func(int a) => a;
  void Test(string dmdtype)
  {
    try
    {
      Switches.SetSwitchValue(Switches.DMDType, dmdtype);
      var newfunc = dmd.Generate().CreateDelegate<Func<int, int, int>>();
      Assert.Equal(3, newfunc(1, 2));
    }
    finally
    {
      Switches.ClearSwitchValue(Switches.DMDType);
    }
  }
}
```

xADDBx:
This adds the test from #289 and forces it to run sequentially (as otherwise it will cause other tests to fail, probably because SetSwitchValue)

Further it removes (comments out) the branches responsible for copying the parameters from the original in both the MethodBuilder and the DynamicMethod generators. I'm not sure if this should be left as is. It does fix the test (and doesn't break any of the old ones). I assume this branch existed for performance reasons? As far as I can see, the Cecil generator already ignores the original method for the sake of parameters.

Note: I suggest hiding whitespace when looking at the diff

### DIFF
--- a/src/MonoMod.UnitTest/Github/Issue282.cs
+++ b/src/MonoMod.UnitTest/Github/Issue282.cs
@@ -1,0 +1,54 @@
+using MonoMod.Cil;
+using MonoMod.Utils;
+using MonoMod;
+using Mono.Cecil.Cil;
+using System;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MonoMod.UnitTest.Github
+{
+    public class Issue282 : TestBase
+    {
+        public Issue282(ITestOutputHelper helper) : base(helper)
+        {
+        }
+
+        [Fact]
+        public void DMDGenerateDoesNotOverrideParameters()
+        {
+            using var dmd = new DynamicMethodDefinition(((Delegate)func).Method);
+            // modify to func(a, b) => a + b;
+            dmd.Definition.Parameters.Add(new(dmd.Module.TypeSystem.Int32));
+            using (var il = new ILContext(dmd.Definition))
+            {
+                var c = new ILCursor(il);
+                c.Index += 1;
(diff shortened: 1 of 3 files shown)


nike4613:
Is there a reasonable-ish way to avoid recreating if we know nothing has changed?

xADDBx:
I don't know how it'd be possible to figure out the "nothing changed" without having to resolve the parameters anyway. And once they're resolved it doesn't make much of a difference anymore.

I think it'd be necessary to compare the length of the parameters and then whether all parameters have the same type and modifiers.

PR #306, opened 2026-05-29
Fix Rel32Ind64 detour issue in Mono to prevent patch loss under Proton

Kuinox:
Hello, 
So I had an issue making rimworld work with some mods on linux.  
I made my AI agents work on this and to fix at the root cause.  

With this patch the game play fine.  
I'm opening this PR so you are aware of this bug & patch, I understand if you don't want to accept it.  

I'm a C# dev but I don't know stuff that low level, I'm open to fix any PR feedback & learn.

### DIFF
--- a/src/MonoMod.UnitTest/Core/MonoDetourPointerCellTest.cs
+++ b/src/MonoMod.UnitTest/Core/MonoDetourPointerCellTest.cs
@@ -1,0 +1,80 @@
+using MonoMod.Core.Platforms;
+using MonoMod.Utils;
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace MonoMod.UnitTest.Core
+{
+    /// <summary>
+    /// Regression test for the Unity Mono + Proton silent patch-loss bug.
+    ///
+    /// On x86-64, <c>Rel32Ind64</c> places its pointer cell in a separately-allocated page.
+    /// Under Unity Mono + Proton, Wine's <c>VirtualQuery</c> reports Mono's JIT memory
+    /// reservation as <c>MEM_FREE</c> because Mono reserved it via a direct Linux
+    /// <c>mmap</c> syscall that bypasses Wine's virtual-memory tracking.  MonoMod's
+    /// allocator therefore places the cell inside that reservation.  When Mono's JIT later
+    /// commits or decommits pages in the same range the cell is overwritten with zeros, so
+    /// the indirect <c>JMP [cell]</c> jumps to null and Mono raises
+    /// <c>NullReferenceException</c> at the patched method's entry point (+0x00000).
+    /// </summary>
+    public sealed class MonoDetourPointerCellTest : TestBase
+    {
+        public MonoDetourPointerCellTest(ITestOutputHelper helper) : base(helper) { }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static void AnchorMethod() { GC.KeepAlive(null); }
+
+        [Fact]
+        public void Rel32Ind64_pointer_cell_is_silently_zeroed_when_page_is_reclaimed()
+        {
+            if (PlatformDetection.Architecture != ArchitectureKind.x86_64)
+                return;
+
+            var anchorMethod = typeof(MonoDetourPointerCellTest)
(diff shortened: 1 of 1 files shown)


src/MonoMod.Core/Platforms/Architectures/x86_64Arch.cs
```diff
@@ -348,10 +348,17 @@ public override int DoRetarget(NativeDetourInfo origInfo, IntPtr to, Span<byte>
...
 
         private readonly ISystem system;
-
-        public x86_64Arch(ISystem system)
+        // When true, skip Rel32Ind64 (which uses a separately allocated pointer cell) and fall
+        // through to Abs64 (inline pointer cell). This is required on runtimes where the JIT
+        // memory manager can silently overwrite independently-allocated pages — in particular,
+        // Unity Mono running under Proton, where Rel32Ind64 cells are invalidated during assembly
+        // load, causing detours to stop redirecting without any error.
+        private readonly bool preferInlineDetour;
```
nike4613: I very strongly dislike having this here. This feels like either a) a memory allocator issue (in which case it should be fixed in the allocator) or b) a runtime quirk which should have the runtime telling the arch impl *at detour create time* that it wants to avoid indirect detours.

src/MonoMod.UnitTest/Core/MonoDetourPointerCellTest.cs
```diff
@@ -0,0 +1,224 @@
...
+            finally
+            {
+                info.InternalData?.Dispose();
+            }
+        }
+
+        [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.NoInlining)]
+        private static void PlaceholderForAddressAnchor() { }
+    }
+}
```
nike4613, line 80: This whole test feels much too narrow. In general, MonoMod does *not* care about the specific behavior of the internals; instead, it cares about the result. Mocking (as is done in here) is basically guaranteed to hide issues. 

Basically all other tests exhibit some scenario where, without some particular fix, detouring *doesn't work* (whether because the process crashes, the detour is unstable, gets the wrong values, etc). Then, with the fix applied, it *does* work, and the test passes. With that pattern, it also means that (mostly) tests aren't skipped based on the host, because we want those scenarios to work for other hosts too!

nike4613 (changes requested):
I am unconvinced that this fix actually fixes the described issue. Please include a test which actually exhibits the issue when run under a relevant Mono build.

If Mono really is silently overwriting the indirection target, I would expect crashes (and not spurious ones!). That also would suggest that Mono has a potentially extremely problematic memory allocator (not out of the question, but I'd want pointers into Mono source for this behavior, or at least an application which semi-reliably reproduces).

What I find more likely is Mono seeing the indirect jump and assuming it's one of its own stubs; this would cause crashes to not happen, but still cause the detours to silently break. Again, however, I want an example which demonstrates the issue consistently, or else source references.

Kuinox:
So I hooked a debugger to understand what is happening.
It looks like when MonoMod scan for a free page, it select a PROT_NONE page allocated by mono gc, because wine doesn't do some ownership checks.

I only managed to reproduce that using mono shipped by unity, not the regular mono.  
I'm pushing a better test soon.

Kuinox:
Sorry I don't think I can help more, I don't know enough the low level stuff do distinguish when the ai is bullshiting or not here.  

I'll keep trying a bit then i'll probably give up.

xADDBx:
While I can't comment on the details; a week ago someone did report an issue that *seemed* MonoMod related that happened with a Unity Mono game on Proton. The StackTrace did look like this
```
NullReferenceException: Object reference not set to an instance of an object
  at XXX.XXX() <0x71bb2a20 + 0x00000> in <90a9869c27924c7bbfb75a8b33da7c82>:0 
```
I didn't investigate myself since
1. It disappeared after the user switched to another Proton version
2. It reminded me of the Bottom-Up ASLR issue I've seen on Windows which was pretty impossible to reproduce properly and

Kuinox:
From the debugger I can see something is zeroing out the detour fonction, but I didnt managed to find what yet.

xADDBx:
@Kuinox do you think the issue addressed in #308 is the same one as this?

Kuinox:
Yes it does look like.

PR #307, opened 2026-06-11
Add support for .NET 11

zznty:
there wasnt much change to jit besides usual guid and vtable adjustments, would probably stay draft till GA


fiinix00:
Tested patch on **Release Candidate 1** (NET 11.0.0-rc.1.26425.128)
In combination with [LibHarmony](https://github.com/pardeike/harmony) (That depends on MonoMod)
Works well, Windows 10 x64

nike4613:
MacOS seems to be failing pretty consistently under Rosetta. Unfortunately crash dumps from there don't work properly... This weekend I'll try to debug it.

src/MonoMod.Core/Platforms/Memory/QueryingPagedMemoryAllocator.cs
```diff
@@ -157,15 +240,26 @@ private unsafe bool TryAllocNewPage(PositionedAllocationRequest request, ref nin
...
+                }
                 else
-                    page = baseAddr - PageSize;
+                {
+                    // downward the query only reports the *next* region, never where free space begins below, so
+                    // descend geometrically: stepping one page across a large unusable gap costs hundreds of
+                    // thousands of OS calls, and a fixed stride either degrades the same way or skips the range
+                    page = baseAddr - pageStep;
+                    if (pageStep < MaxPageStep)
+                        pageStep *= 2;
```
nike4613: This scheme seems like it can cause allocation to fail (because we skipped valid pages) when it is serviceable. If we do something like this in the allocator, I'd want it to keep track of the ranges it skipped and use query information to scan through the actual free pages in that region. That said, I'm not particularly convinced this is necessary in any case.

src/MonoMod.Core/Platforms/Systems/MacOSSystem.cs
```diff
@@ -477,6 +477,31 @@ public override unsafe bool TryAllocatePage(IntPtr pageAddr, nint size, bool exe
...
+            public override unsafe bool TryAllocatePageNear(IntPtr hint, nint size, bool executable, out IntPtr allocated)
+            {
+                // A non-fixed mapping treats the address as a hint: the kernel places the page at the hint when it is
+                // free and reports where it actually landed (measured on a translated x64 process: 200/200 hints
+                // honoured, ~135us per probe, against seconds for a region-by-region search). The caller validates
+                // the returned address against its bounds, and executable pages keep using the fixed path above.
+                // arm64 requires MAP_JIT for writable-executable mappings and hands that path to the fixed
+                // allocator; on x86_64 the hint path can place either kind of page.
+                if (executable && PlatformDetection.Architecture == ArchitectureKind.Arm64)
+                    return TryAllocatePage(hint, size, executable, out allocated);
```
nike4613: If we can just defer and get correct behavior, what even is the point of this method? If this hinting is possible, why are we not using it as a first-attempt always, rather than only in certain cases?

src/MonoMod.Core/Platforms/Memory/QueryingPagedMemoryAllocator.cs
```diff
@@ -39,6 +39,17 @@ public abstract class QueryingMemoryPageAllocatorBase
...
+        /// <summary>
+        /// Tries to allocate a page at <paramref name="hint"/>, or as close to it as the OS is willing to place it.
+        /// </summary>
+        /// <remarks>
+        /// Platforms which can place a mapping near an address without first proving that address is free implement
+        /// this; the range search uses it to avoid walking the region map (which costs one or two OS calls per region).
+        /// Implementations must report the address they actually used, which the caller validates against its bounds.
+        /// </remarks>
+        public virtual bool TryAllocatePageNear(IntPtr hint, nint size, bool executable, out IntPtr allocated)
+            => TryAllocatePage(hint, size, executable, out allocated);
```
nike4613: What is the value of having this as a separate method? The contract here seems to be approximately the same as `TryAllocatePage`.

nike4613 (changes requested):
I don't think allocator changes here are the fix? Still getting reliable timeouts it seems, so even if allocator perf is improved with this (which I believe, though I have some conceptual issues with the strategy), not in a way which affects the hangs. I'd rather any allocator changes be done in a separate PR so it can be reviewed in isolation.

PR #313, opened 2026-08-03
Add ILHookTransaction to batch per-target IL chain rebuilds

std-microblock:
Introduces an opt-in startup transaction that defers ILHook.Apply until a group of hooks can be committed together. Hooks targeting the same method are inserted into the hook graph in a single batch, so the source IL is cloned, every manipulator is replayed and the DynamicMethod is generated and JIT-compiled only once per target method instead of once per hook.

- ILHook.Apply/Undo route through the transaction while it is active and IsApplied reflects the pending state.
- DetourManager.AddILHooksBatch inserts a whole target group atomically and rolls the graph back if chain preparation fails.
- Flush orders hooks by a stable (order, sequence) key and can serialize manipulators per owner key while running clone/generate/JIT outside the owner gate, preserving both per-mod and per-target hook ordering.

Outside an active transaction, ILHook behavior is unchanged.

Usage example: https://github.com/EverestAPI/Everest/pull/1153

Note: This pull request is written under the assistance of AI

### DIFF
--- a/src/MonoMod.RuntimeDetour/ILHook.cs
+++ b/src/MonoMod.RuntimeDetour/ILHook.cs
@@ -5,6 +5,7 @@ using MonoMod.Utils;
 using System;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Threading;
 
 namespace MonoMod.RuntimeDetour
 {
@@ -158,6 +159,13 @@ namespace MonoMod.RuntimeDetour
 
         private readonly DetourManager.ManagedDetourState state;
         private readonly DetourManager.SingleILHookState hook;
+        private int transactionPending;
+
+        internal DetourManager.ManagedDetourState ManagedState => state;
+        internal DetourManager.SingleILHookState HookState => hook;
+
+        internal void SetTransactionPending(bool pending)
+            => Volatile.Write(ref transactionPending, pending ? 1 : 0);
 
         /// <summary>
         /// Constructs an <see cref="ILHook"/> for the provided method using the provided manipulator and <see cref="DetourConfig"/>
@@ -197,7 +205,7 @@ namespace MonoMod.RuntimeDetour
         /// <summary>
         /// Gets whether or not this <see cref="ILHook"/> is applied.
         /// </summary>
-        public bool IsApplied => hook.IsApplied;
+        public bool IsApplied => hook.IsApplied || Volatile.Read(ref transactionPending) != 0;
         /// <summary>
         /// Gets the <see cref="ILHookInfo"/> for this <see cref="ILHook"/>.
         /// </summary>
@@ -223,6 +231,8 @@ namespace MonoMod.RuntimeDetour
                 if (IsApplied)
                     return;
                 MMDbgLog.Trace($"Applying ILHook for {Method}");
+                if (ILHookTransaction.TryQueue(this))
+                    return;
                 state.AddILHook(hook, !lockTaken);
             }
             finally
@@ -246,6 +256,8 @@ namespace MonoMod.RuntimeDetour
                 if (!IsApplied)
                     return;
                 MMDbgLog.Trace($"Undoing ILHook for {Method}");
+                if (ILHookTransaction.TryCancel(this))
+                    return;
                 state.RemoveILHook(hook, !lockTaken);
             }
             finally
--- a/src/MonoMod.RuntimeDetour/ILHookTransaction.cs
+++ b/src/MonoMod.RuntimeDetour/ILHookTransaction.cs
@@ -1,0 +1,301 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace MonoMod.RuntimeDetour
+{
+    /// <summary>
+    /// Defers a group of <see cref="ILHook.Apply"/> operations and commits all
+    /// hooks targeting the same method with a single IL-chain rebuild.
+    /// </summary>
+    /// <remarks>
+    /// This is intended for controlled startup phases where many independent
+    /// components install hooks before the target methods can run. Outside an
+    /// active transaction, ILHook behavior is unchanged.
+    /// </remarks>
+    public sealed class ILHookTransaction : IDisposable
+    {
+        private sealed class OrderContext
+        {
+            public readonly long Order;
+            public long Sequence;
+
+            public OrderContext(long order)
+            {
+                Order = order;
+            }
+        }
+
+        private sealed class PendingHook
+        {
+            public readonly ILHook Hook;
+            public readonly long Order;
+            public readonly long LocalSequence;
+            public readonly long GlobalSequence;
+
+            public PendingHook(ILHook hook, long order, long localSequence, long globalSequence)
+            {
+                Hook = hook;
+                Order = order;
+                LocalSequence = localSequence;
+                GlobalSequence = globalSequence;
+            }
+        }
+
+        private sealed class Scope : IDisposable
+        {
+            private readonly OrderContext? previous;
+            private int disposed;
+
+            public Scope(OrderContext? previous)
+            {
+                this.previous = previous;
+            }
+
+            public void Dispose()
+            {
+                if (Interlocked.Exchange(ref disposed, 1) == 0)
+                    currentOrder.Value = previous;
+            }
+        }
+
+        private sealed class MonitorScope : IDisposable
+        {
+            private object? gate;
+
+            public MonitorScope(object gate)
+            {
+                this.gate = gate;
+                Monitor.Enter(gate);
+            }
+
+            public void Dispose()
+            {
+                var value = Interlocked.Exchange(ref gate, null);
+                if (value is not null)
+                    Monitor.Exit(value);
+            }
+        }
+
+        private static readonly object transactionLock = new();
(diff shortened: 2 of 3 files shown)


src/MonoMod.RuntimeDetour/ILHookTransaction.cs
```diff
@@ -0,0 +1,301 @@
...
+                var value = Interlocked.Exchange(ref gate, null);
+                if (value is not null)
+                    Monitor.Exit(value);
+            }
+        }
+
+        private static readonly object transactionLock = new();
+        private static readonly AsyncLocal<OrderContext?> currentOrder = new();
+        private static readonly ConcurrentDictionary<ILHook, ILHookTransaction> pendingOwners = new();
+        private static ILHookTransaction? activeTransaction;
```
nike4613, line 86: So there can only ever be one active transaction **globally**, across all threads? That seems like a huge problem, especially considering there's no exposed way to synchronize against it; you can only try to start a transaction and see if it throws.

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// Associates subsequently queued hooks on the current execution context
+        /// with a stable ordering key. Hooks with the same key preserve call order.
+        /// </summary>
+        public static IDisposable EnterOrder(long order)
+        {
+            var previous = currentOrder.Value;
+            currentOrder.Value = new OrderContext(order);
+            return new Scope(previous);
+        }
+
```
nike4613, line 124: Why do we have this ordering mechanism? Hooks and ILHooks already have ordering as part of their `DetourConfig`; the order of `Apply()` only matters when that's tied, and even then we make no guarantees. It seems strange, then, to try to expose that call ordering here as if its a contract that matters.

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
```diff
@@ -0,0 +1,301 @@
...
+        private IDisposable EnterManipulatorGate(MonoMod.Cil.ILContext.Manipulator manipulator,
+            Func<MonoMod.Cil.ILContext.Manipulator, object?>? ownerSelector)
+        {
+            var owner = ownerSelector?.Invoke(manipulator) ?? manipulator.Method.DeclaringType?.Assembly;
+            var gate = owner is null ? unknownOwnerGate : ownerGates.GetOrAdd(owner, static _ => new object());
+            return new MonitorScope(gate);
+        }
+
```
nike4613, line 278: This will cause the hooking process to take locks while already under other locks. This seems like it would be prone to deadlocks. Do we have any guarantee that it will not? (Why is this even necessary?)

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
```diff
@@ -0,0 +1,301 @@
...
+                {
+                    foreach (var group in groups)
+                        CommitGroup(group);
+                }
+                else
+                {
+                    Parallel.ForEach(groups, new ParallelOptions {
+                        MaxDegreeOfParallelism = maxDegreeOfParallelism
+                    }, CommitGroup);
+                }
```
nike4613, line 253: Why do we want parallel commit? This seems like it could only cause problems.

src/MonoMod.RuntimeDetour/ILHook.cs
```diff
@@ -223,6 +231,8 @@ public void Apply()
                 if (IsApplied)
                     return;
                 MMDbgLog.Trace($"Applying ILHook for {Method}");
+                if (ILHookTransaction.TryQueue(this))
+                    return;
                 state.AddILHook(hook, !lockTaken);
```
nike4613, line 236: Why are we doing implicit global-state transactions, instead of an explicit object which is given the hooks to apply? This seems like it would be very easy to get unintended and possibly wrong behavior, on top of being unintuitive.

nike4613 (changes requested):
Build is failing. Did you even try to build locally?

std-microblock:
> Build is failing. Did you even try to build locally?

hm, works on my machine, why

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
std-microblock: In downstream PR I used this to ensure the hook applying order is the same as being applied serially. Yea that might not be guaranteed by monomod but I wanted to keep changes in behavior smaller to avoid having extra issues.

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
std-microblock: That's because I'm avoiding changing the interface of ILHook so that the Mod authors don't have to change their code to adopt to the feature. Would there be another better way? Thread local is also unacceptable because I'm parallelizing the mod main to get performance gains

std-microblock: Do need more APIs to query transaction state and force flush indeed

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
std-microblock: To make it faster..? In my tests it speed the commiting process up for like 2x under 2 parallelism and 2.5x under 4 parallelism, like 4 seconds to 2 seconds. Is it better to simply expose related api to user of the lib rather than do it here

src/MonoMod.RuntimeDetour/ILHook.cs
std-microblock: Same reason as https://github.com/MonoMod/MonoMod/pull/313#discussion_r3709238327

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
std-microblock: This lock here is also for parallelizing the commiting process. It's used to keep commiting from the same assemblies serialed as we don't know if the hook callback the user wrotes is thread safe or not. If parallelism is removed, this can also be removed. https://github.com/MonoMod/MonoMod/pull/313#discussion_r3709251322

std-microblock:
> > Build is failing. Did you even try to build locally?
> 
> hm, works on my machine, why

oh that's bcuz the ci is using net35 and net452, time to upgrade xD

anyway if we decide to remove the parallelism part there should be no problem

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
nike4613: I don't think we'd need to change the public API of ILHook. Consider:

```cs
var tns = ILHookTransaction.Begin();
tns.Apply(new ILHook(..., applyByDefault: false));
// ...
tns.Commit();
```

nike4613:
> > > Build is failing. Did you even try to build locally?
> > 
> > 
> > hm, works on my machine, why
> 
> oh that's bcuz the ci is using net35 and net452, time to upgrade xD
> 
> anyway if we decide to remove the parallelism part there should be no problem

The normal build process builds all target frameworks. CI just does `dotnet build`.

src/MonoMod.RuntimeDetour/ILHookTransaction.cs
std-microblock: > I don't think we'd need to change the public API of ILHook. Consider:
> 
> ```cs
> var tns = ILHookTransaction.Begin();
> tns.Apply(new ILHook(..., applyByDefault: false));
> // ...
> tns.Commit();
> ```

Many mod still uses `new ILHook(...).apply()` way and this will also force them to change. I have an idea suddenly tho, is it better for us to pass transaction context through a global threadlocal as a hidden context? Like

```cs
var tns = ILHookTransaction.Begin();
new ILHook(...).apply();
tns.Commit();
```

std-microblock: hmm, in the view of a universal hooking framework it do sounds strange to change the behaviors of `.apply()`, probably add a explicit option `deferAllApplies` in ILHookTransaction

nike4613: Is the goal for this to be transparent to mod authors? That seems potentially dangerous, as it meaningfully changes behavior in a way that existing ambient state doesn't. (The existing DetourContext deals with exactly this kind of thing for automatically associated e.g. a mod ID with detours)

std-microblock: > Is the goal for this to be transparent to mod authors? 

yeah I designed it with this in mind. The reason why I modified monomod is actually to allow downstream modloaders(Everest for example) to speed up mod loading process by batching hook applies and avoid unnecessary decompilation/compilation, so it's not designed for a single mod to use(that won't have much effect) and I do want it to be transparent to mod authors

> That seems potentially dangerous, as it meaningfully changes behavior in a way that existing ambient state doesn't. (The existing DetourContext deals with exactly this kind of thing for automatically associated e.g. a mod ID with detours)

What behavior? I actually have implemented a mechanism where calling a hooked function results in a instant flush if it's dirty, and as this is implemented the only observable side effect of enabling deferAllApplies is throwing exceptions in ilhook callbacks may not be catched by the try-catch block outside of hook.apply(); but ye that's a big chanhe and I'm still considering how to deal with that.

PR #319, opened 2026-08-20
Fix #318 (Wrong method entry point for NGEN-ed method pre 4.8)

### LINKED ISSUE #318: Wrong method entry point on .NET Framework 4.6.2 / x86: non-vtable slot holds a relative pointer for NGEN'd methods
## Summary

On **.NET Framework 4.6.2 / x86**, `FxCLR4Runtime.GetMethodBodyPtr` returns a wrong address for NGEN'd methods that have `HasNonVtableSlot == true`. The returned value is the **relative** pointer stored in the non-vtable slot, used as if it were an **absolute** pointer.

The consequence is either an `AccessViolationException` that terminates the process, or — worse — a **successful detour write into unrelated memory** (in our case, into the memory-mapped `C:\Windows\System32\locale.nls`), which is silent memory corruption.

This appears to be exactly what the two existing `TODO`s in the code are about:

- [`FxCLR4Runtime.cs#L39`](https://github.com/MonoMod/MonoMod/blob/main/src/MonoMod.Core/Platforms/Runtimes/FxCLR4Runtime.cs#L39) — `// TODO: check to make sure we're running 4.8 before using this`
- [`Fx.V48.cs#L185`](https://github.com/MonoMod/MonoMod/blob/main/src/MonoMod.Core/Interop/Fx.V48.cs#L185) — `// TODO: figure out why Process.Start on x86 specifically`

Still reproducible on `main` (`1245fef6` only touched the ARM64 ABI setup, not this path).

## Environment

| | |
|---|---|
| OS | Windows 7 SP1 (6.1.7601), freshly installed |
| Runtime | .NET Framework **4.6.2** (registry `Release=394806`, `FrameworkDescription = ".NET Framework 4.6.1590.0"`) |
| Architecture | **x86** (`Prefer32Bit=true`) |
| Lib.Harmony | 2.4.2 (MonoMod.Core 1.3.3) |
| Detected runtime | `Framework 4.6.1590.0 using Framework corelib` (from MonoMod's own log) |

## Reproduction

Patching `System.Net.WebRequest.Create(string)` (or `Create(Uri)` / `CreateDefault(Uri)`) with Harmony:

```csharp
var harmony = new Harmony("repro");
var target = typeof(WebRequest).GetMethod("Create", BindingFlags.Public | BindingFlags.Static,
    null, new[] { typeof(string) }, null);
harmony.Patch(target, postfix: new HarmonyMethod(somePostfix));
```

```
System.AccessViolationException
   at MonoMod.Core.Platforms.Architectures.x86Shared.TryRel32Detour(IntPtr, IntPtr, Int32, NativeDetourInfo ByRef)
   at MonoMod.Core.Platforms.Architectures.x86Arch.ComputeDetourInfo(IntPtr, IntPtr, Int32)
   at MonoMod.Core.Platforms.PlatformTriple.CreateSimpleDetour(IntPtr, IntPtr, Int32, IntPtr)
   ...
   at HarmonyLib.PatchFunctions.UpdateWrapper(MethodBase, PatchInfo)
```

MonoMod's own log confirms it is about to detour a bogus address:

```
Info: Detected runtime: Framework 4.6.1590.0 using Framework corelib
Trace: Creating simple detour 0x000000000017702c => 0x0000000001fa0100
```

`0x0017702c` is not code — it lands inside `locale.nls`.

## Root cause

`GetMethodBodyPtr` parses the `MethodDesc` using the **4.8** layout unconditionally:

```csharp
// FxCLR4Runtime.cs:39-48
// TODO: check to make sure we're running 4.8 before using this
private unsafe IntPtr GetMethodBodyPtr(MethodBase method, RuntimeMethodHandle handle)
{
    var md = (V48.MethodDesc*)handle.Value;
    md = V48.MethodDesc.FindTightlyBoundWrappedMethodDesc(md);
    var ptr = (IntPtr)md->GetNativeCode();
    return ptr;
}
```

For the methods above, the flags read as `HasNativeCodeSlot=false`, `HasStableEntryPoint=true`, `HasPrecode=false`, so `GetNativeCode()` falls through to `GetStableEntryPoint()` → `GetMethodEntryPoint()`, which reads the non-vtable slot at `this + 8` and **dereferences it as an absolute pointer**:

```csharp
// Fx.V48.cs:104-117
public void* GetMethodEntryPoint()
{
    if (HasNonVtableSlot)
    {
        var size = GetBaseSize();
        var pSlot = ((byte*)Unsafe.AsPointer(ref this)) + size;
        return *(void**)pSlot;          // <-- treated as absolute
    }
    return MethodTable->GetSlot(SlotNumber);
}
```

On 4.6.2, for NGEN'd methods, that slot holds a **relative** pointer. Adding the slot's own address recovers the correct entry point exactly. Three independent methods, all matching:

| Method | slot addr | slot value | slot + addr | `handle.GetFunctionPointer()` |
|---|---|---|---|---|
| `WebRequest.Create(string)` | `0x6cb607f8` | `0x0017702c` | `0x6ccd7824` | `0x6ccd7824` ✅ |
| `WebRequest.Create(Uri)` | `0x6cb60804` | `0x00183cac` | `0x6cce44b0` | `0x6cce44b0` ✅ |
| `WebRequest.CreateDefault(Uri)` | `0x6cb60810` | `0x005df854` | `0x6d140064` | `0x6d140064` ✅ |

Raw `MethodDesc` dump for the first one (`handle.Value = 0x6cb607f0`):

```
85 8e 10 31 | 44 00 28 20 | 2c 70 17 00 | 86 8e 13 31
                            ^^^^^^^^^^^ +0x08 = 0x0017702c (relative)
m_bFlags2 = 0x31 -> HasStableEntryPoint
m_wFlags  = 0x2028 -> HasNonVtableSlot | Static
```

### Why it is version- and NGEN-specific

- On **4.8.1 / x86** the same slot contains an **absolute** pointer, so the current code is correct there. Dump of the same method on 4.8.1: `db 8e 0b 31 | 44 00 28 20 | bc 64 40 70` → `+0x08 = 0x704064bc`, which *is* the entry point.
- A **non-NGEN'd (JIT-only)** method on the very same 4.6.2 machine also stores an absolute pointer at `+0x08` and is handled correctly. So the problem is specific to NGEN'd images (`*.ni.dll`).
- `HttpClientHandler.InitializeWebRequest` has `HasNonVtableSlot=false` and goes through `MethodTable->GetSlot()`, so it patches fine on 4.6.2 — which is why the failure looks selective.

This also matches the observation in pardeike/Harmony#21 back in 2017 (patching NGEN'd GAC assemblies fails, non-framework classes are fine), although that issue predates MonoMod and was closed with a `VirtualProtect` change.

## Workaround

Since `FxCoreBaseRuntime.GetMethodEntryPoint` (the base implementation, also used by `FxCLR2Runtime`) already does the right thing:

```csharp
// FxCoreBaseRuntime.cs:477
return handle.GetFunctionPointer();
```

…forcing `GetMethodBodyPtr` to return `handle.GetFunctionPointer()` fixes it. `GetMethodEntryPoint` already calls `TryInvokeBclCompileMethod` beforehand, so the method is guaranteed to be compiled at that point and the returned pointer is the stable entry point rather than a precode.

Verified on the affected machine (3 runs, isolated processes):

| Configuration | patch result | postfix actually invoked |
|---|---|---|
| unmodified 2.4.2 | 1 ok / 3 `AccessViolationException` | **0** |
| `GetMethodBodyPtr` → `GetFunctionPointer()` | **4 ok / 0 failed** | **3** |
| slot parsed as relative pointer | 4 ok / 0 failed | 3 |

The last row is just a cross-check that the "relative pointer" reading is correct.

Note that the `postfix actually invoked` column matters: with the unmodified version, one of our earlier runs had `harmony.Patch` **report success** while having written `e9 <rel32>` into `locale.nls`. Only checking for absence of exceptions is not sufficient to detect this bug.

## Suggested fix

A few options, in the maintainers' hands:

1. Gate the `V48.MethodDesc` path on an actual 4.8 check and fall back to the base implementation otherwise — this is what the existing `TODO` suggests.
2. Determine whether the non-vtable slot holds a relative or absolute pointer (the `FIXUP_LIST_MASK`-style bit trick already used in `GetNativeCode`'s `HasNativeCodeSlot` branch may be applicable here too) instead of assuming absolute.
3. Sanity-check the computed address before detouring (e.g. verify it is readable and lies within an executable region) so that a mis-parse degrades into an exception instead of silent memory corruption.

Happy to run further diagnostics on the affected machine if that helps — we still have access to it and have tooling in place to dump `MethodDesc` contents and compare address computations.

xADDBx:
The issue seems to go back to an update in .NET Framework. .NET Framework 4.8 contains the following in its [changelog](https://learn.microsoft.com/en-us/dotnet/framework/whats-new/#common-language-runtime)
> NGEN improvements [...]

I looked at .NET 4.7.2 and .NET 4.8 binaries and confirmed that the relevant behavior did change in between the two of them.

There is already a test for this behavior in https://github.com/MonoMod/MonoMod/blob/reorganize/src/MonoMod.UnitTest/RuntimeDetour/DetourExtTest.cs#L205. None of the existing runners catch this though because they are based on modern windows which ships .NET Framework 4.8.1. I did test on older runtimes and manually installed .NET 4.6.2 and .NET 4.7.1 and could reproduce the issue with the specified test.

The fix is based on the behavior in [CoreCLR](https://github.com/dotnet/coreclr/blob/v1.0.0/src/vm/method.cpp#L570)
```cpp
if (HasNonVtableSlot())
{
    SIZE_T size = GetBaseSize();

    TADDR pSlot = dac_cast<TADDR>(this) + size;

    return IsZapped() ? NonVtableSlot::GetValueAtPtr(pSlot) : *PTR_PCODE(pSlot);
}
```

I implemented this as a hardcoded condition. A better place for this might be a new `RuntimeFeature` flag. I didn't do that for now since the condition is trivial.

Fixes #318

### DIFF
--- a/src/MonoMod.Core/Interop/Fx.V48.cs
+++ b/src/MonoMod.Core/Interop/Fx.V48.cs
@@ -107,6 +107,13 @@ namespace MonoMod.Core.Interop
                     {
                         var size = GetBaseSize();
                         var pSlot = ((byte*)Unsafe.AsPointer(ref this)) + size;
+                        // In runtimes before .NET Framework 4.8, NGEN causes slots to contain relative offsets
+                        // https://github.com/MonoMod/MonoMod/issues/318
+                        if (PlatformDetection.RuntimeVersion < new Version(4, 8) && MethodDescChunk->m_flagsAndTokenRange.Has(V48.MethodDescChunk.Flags.IsZapped))
+                        {
+                            var relativeEntryPoint = ((RelativePointer*)pSlot)->Value;
+                            return relativeEntryPoint;
+                        }
 
                         return *(void**)pSlot;
                     }
--- a/src/MonoMod.Core/Platforms/Runtimes/FxCLR4Runtime.cs
+++ b/src/MonoMod.Core/Platforms/Runtimes/FxCLR4Runtime.cs
@@ -36,7 +36,6 @@ namespace MonoMod.Core.Platforms.Runtimes
         public override RuntimeFeature Features
             => base.Features & ~RuntimeFeature.RequiresBodyThunkWalking;
 
-        // TODO: check to make sure we're running 4.8 before using this
         private unsafe IntPtr GetMethodBodyPtr(MethodBase method, RuntimeMethodHandle handle)
         {
             var md = (V48.MethodDesc*)handle.Value;


xADDBx:
I just noticed that PlatformDetection.RuntimeVersion < new Version(4, 8) is too broad.
Iirc things like .NET Core can mess up a lot with versioning; so there should either also be a check for .NET Framework *or* it should be implemented as a RuntimeFeature.

nike4613:
That codepath is only followed on FX, so the fix should be fine?

xADDBx:
> That codepath is only followed on FX, so the fix should be fine?

That explains why .NET Core didn't fail.

nike4613:
Is there a reasonable way we can make some runners that test these older netfx, or is that just totally hopeless? My memory is that its at least not easy.

xADDBx:
I only took a glance before I PR-ed but didn't see a way without a custom runner.

A cursed idea would be a linux container using a windows vm maybe?

nike4613:
It looks like it might be possible:
https://mcr.microsoft.com/en-us/artifact/mar/dotnet/framework/runtime/tag/4.6.2
https://mcr.microsoft.com/en-us/artifact/mar/dotnet/framework/runtime/tag/4.7.2

xADDBx:
Hmm. I'm not sure if we can use those on a normal GitHub hosted runner

nike4613 (approved):

nike4613:
https://github.com/actions/runner/issues/904#issuecomment-3376129108 suggests it is doable, and other parts of the thread suggest using Actions to run a custom build of the runner with native support enabled. So I think this should be doable, if not easy. I may mess with it this weekend.

Anyway, I have no problem with this change, assuming it works for the scenario that prompted this.

xADDBx:
We can keep the PR open until we either have the runner or the issue creator confirms the fix I guess

xADDBx:
> [actions/runner#904 (comment)](https://github.com/actions/runner/issues/904#issuecomment-3376129108) suggests it is doable, and other parts of the thread suggest using Actions to run a custom build of the runner with native support enabled. So I think this should be doable, if not easy. I may mess with it this weekend.

I did a quick check; mcr.microsoft.com/dotnet/framework/runtime:4.6.2 seems to at least load; but mscorlib reports release `461814` which is .NET Framework 4.7.2?

Which still works if that runs; but it's pretty weird.

xADDBx:
I made a basic workflow based on the suggested container (just to verify this fix):

[Workflow without fix](https://github.com/xADDBx/MonoMod/actions/runs/32451305005/job/96681159133) (TestDetoursExt failing)
[Workflow with fix](https://github.com/xADDBx/MonoMod/actions/runs/32451950203/job/96688308802) (TestDetoursExt succeeding)

